PR #1, opened 2025-04-17
Project setup

SandipBajracharya:
- [X] ESLint & Prettier + ESLint Prettier Plugin
- [X] Husky & Lint staged on pre-commit
- [X] Lint workflow
- [X] PULL_REQUEST_TEMPLATE.md
- [X] CopilotAPI.ts file
- [X] Drizzle setup
- [X] Dependencies version update
- [X] MVC w/ Service architecture for API
- [X] Health check api


eslint.config.mjs
```diff
@@ -0,0 +1,27 @@
...
+    rules: {
+      'prettier/prettier': ['error', { usePrettierrc: true }],
+      '@typescript-eslint/no-unnecessary-type-constraint': 'off',
+      '@typescript-eslint/no-explicit-any': 'off',
+      '@typescript-eslint/no-unsafe-function-type': 'off',
+      '@typescript-eslint/no-unused-vars': ['warn', { argsIgnorePattern: '^_', varsIgnorePattern: '^_' }],
+    },
+  }),
+  {
+    ignores: ['.next', '.cache', 'yarn.lock', 'public', 'node_modules', 'next-env.d.ts', 'next.config.ts', 'yarn.lock'],
```
rrojan, line 23: Looks like `yarn.lock` is repeated twice

eslint.config.mjs
```diff
@@ -0,0 +1,27 @@
...
+const compat = new FlatCompat({
+  // import.meta.dirname is available after Node.js v20.11.0
+  baseDirectory: import.meta.dirname,
+  recommendedConfig: js.configs.recommended,
+})
+
+const eslintConfig = [
+  ...compat.config({
+    extends: ['next/core-web-vitals', 'next/typescript', 'plugin:prettier/recommended'],
+    plugins: ['prettier'],
```
rrojan, line 13: Could you please confirm if we need a typescript plugin here? I think it might be needed.

Here is a lightweight eslint.config.mjs that i used for a recent pet project for reference:

```js
import js from '@eslint/js'
import prettierConfig from 'eslint-config-prettier'
import prettierPlugin from 'eslint-plugin-prettier'
import { defineConfig } from 'eslint/config'
import globals from 'globals'
import tseslint from 'typescript-eslint'

export default defineConfig([
  {
    ignores: ['dist/**', 'tsconfig.json'],
  },
  {
    files: ['**/*.{js,mjs,cjs,ts}'],
    plugins: { js },
    extends: ['js/recommended'],
  },
  {
    files: ['**/*.{js,mjs,cjs,ts}'],
    languageOptions: { globals: globals.node },
  },
  tseslint.configs.recommended,
  prettierConfig,
  {
    plugins: { prettier: prettierPlugin },
    rules: {
      'prettier/prettier': 'error',
    },
  },
  {
    rules: {
      '@typescript-eslint/no-unused-vars': [
        'error',
        {
          varsIgnorePattern: '^_',
          argsIgnorePattern: '^_',
          caughtErrorsIgnorePattern: '^_',
        },
      ],
      'no-console': 'warn', // optional, since i was using pino here
      'prefer-const': 'error',
      'no-var': 'error',
      '@typescript-eslint/explicit-function-return-type': 'warn', // optional as well
      '@typescript-eslint/consistent-type-imports': 'warn',
    },
  },
])

```

.github/workflows/lint.yml
```diff
@@ -0,0 +1,25 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out Git repository
+        uses: actions/checkout@v4
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20.11.1
```
rrojan, line 17: Let's see if we can use v22 instead in `.nvmrc` and actions

.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
+  "singleQuote": true,
+  "printWidth": 125,
```
rrojan, line 3: From our experience in tasks, 125 gets a bit hard to read sometimes. You can consider reducing this but it's optional

drizzle.config.ts
```diff
@@ -0,0 +1,11 @@
+import { DATABASE_URL } from '@/config'
+import { defineConfig } from 'drizzle-kit'
+
+export default defineConfig({
+  schema: './src/db/schema',
+  out: './src/supabase/migrations',
```
rrojan, line 6: Seems like the migrations are placed in `supabase/` and that could conflict with the Supabase CLI, which also generates a supabase/ folder. Can we place them in the db folder itself?

package.json
```diff
@@ -0,0 +1,69 @@
...
+    "lint:check": "eslint \"{src,test}/**/*.{ts,tsx}\"",
+    "prettier:check": "prettier --check \"{src,test}/**/*.{ts,tsx}\"",
+    "lint:fix": "npm run lint:check -- --fix",
+    "prettier:fix": "prettier --write .",
+    "lint-staged": "npx lint-staged",
+    "prepare": "husky"
+  },
+  "dependencies": {
+    "@sentry/nextjs": "^9.13.0",
+    "axios": "^1.8.4",
```
rrojan, line 20: Let's prefer NextJS's fetch over axios since there are a lot of things like cache times, tages, etc that we can only control by using it.

src/app/(home)/home.tsx
```diff
@@ -0,0 +1,5 @@
+'use client'
+
+export default function Home() {
+  return <div className="w-full h-screen flex items-center justify-center">Home page</div>
+}
```
rrojan, line 5: It might be a good idea to name component files Capitalized

src/app/api/core/models/User.model.ts
```diff
@@ -0,0 +1,24 @@
...
+  internalUserId?: string
+
+  // Instantiate a User from a request token & decrypted payload
+  constructor(token: string, tokenPayload: Token) {
+    this.token = token
+    this.internalUserId = tokenPayload.internalUserId
+    this.clientId = tokenPayload.clientId
+    this.companyId = tokenPayload.companyId
+    this.workspaceId = tokenPayload.workspaceId
+  }
```
rrojan, line 21: Let's also add a `role` property here that is Client if `clientId` exists, else IU

src/app/layout.tsx
```diff
@@ -0,0 +1,18 @@
+import './globals.css'
+import type { Metadata } from 'next'
+import { Inter } from 'next/font/google'
+
+const inter = Inter({ subsets: ['latin'] })
+
+export const metadata: Metadata = {
+  title: 'QuickBooks Sync',
+  description: 'QuickBooks Sync Example',
```
rrojan, line 9: Let's remove Example 😄

src/supabase/migrations/0000_demonic_umar.sql
```diff
@@ -0,0 +1,14 @@
...
+	"access_token" varchar NOT NULL,
+	"refresh_token" varchar NOT NULL,
+	"expires_in" integer NOT NULL,
+	"sync_flag" boolean DEFAULT false,
+	"updated_at" timestamp DEFAULT now() NOT NULL,
+	"created_at" timestamp DEFAULT now() NOT NULL,
+	"deleted_at" timestamp
+);
+--> statement-breakpoint
+CREATE UNIQUE INDEX "portal_idx" ON "qb_tokens" USING btree ("portal_id");
```
rrojan, line 14: Can we customize the migration names, seems kind of dark 😄 

image

src/supabase/migrations/0001_large_silhouette.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE "qb_tokens" ALTER COLUMN "portal_id" SET DATA TYPE varchar(255);--> statement-breakpoint
+ALTER TABLE "qb_tokens" ALTER COLUMN "intuit_realm_id" SET DATA TYPE varchar(255);
```
rrojan, line 2: Let's merge these migrations if possible

rrojan (changes requested):

eslint.config.mjs
SandipBajracharya: next/typescript covers the typescript plugins and config. We do not have to define it separately

src/db/schema/qb_tokens.ts
```diff
@@ -0,0 +1,18 @@
...
+export const qb_tokens = table(
+  'qb_tokens',
+  {
+    id: t.integer().primaryKey().generatedAlwaysAsIdentity(),
+    portal_id: t.varchar({ length: 255 }).notNull(),
+    intuit_realm_id: t.varchar({ length: 255 }).notNull(),
+    access_token: t.varchar().notNull(),
+    refresh_token: t.varchar().notNull(),
+    expires_in: t.integer().notNull(),
+    sync_flag: t.boolean().default(false),
```
rrojan, line 14: I believe by setting `snake_case` in `casing`, we can map snake cased SQL column names to camel cased model properties in typescript automatically, so we don't have to use snake case in our typescript key names.

https://orm.drizzle.team/docs/sql-schema-declaration#camel-and-snake-casing

src/db/schema/qb_tokens.ts
```diff
@@ -0,0 +1,18 @@
...
+import * as t from 'drizzle-orm/pg-core'
+
+export const qb_tokens = table(
+  'qb_tokens',
+  {
+    id: t.integer().primaryKey().generatedAlwaysAsIdentity(),
+    portal_id: t.varchar({ length: 255 }).notNull(),
+    intuit_realm_id: t.varchar({ length: 255 }).notNull(),
+    access_token: t.varchar().notNull(),
+    refresh_token: t.varchar().notNull(),
```
rrojan, line 12: Add varchar limits if known

src/db/schema/qb_tokens.ts
```diff
@@ -0,0 +1,18 @@
...
+    expires_in: t.integer().notNull(),
+    sync_flag: t.boolean().default(false),
+    ...timestamps,
+  },
+  (table) => [t.uniqueIndex('portal_idx').on(table.portal_id)],
```
rrojan, line 17:
```suggestion
  (table) => [t.uniqueIndex('unique_qb_tokens_portal_id_idx').on(table.portal_id)],
```

Best to follow a convention so we can understand the index type + table + column from the name itself

rrojan (changes requested):

src/db/schema/qb_tokens.ts
SandipBajracharya: Added for few columns but I think it would be better if we not put constraints on string length db-wise.

rrojan (approved):
LGTM 🏆 
Awesome work!

PR #2, opened 2025-04-23
OUT-1653 | OUT-1654 | OAuth using SDK and token management

SandipBajracharya:
## Changes

- [x] table, schema and migrations
- [x] Internal Copilot auth
- [x] authentication with sdk and create necessary apis (callbacks)
- [x] handle callback and tokens store

## Testing Criteria

QuickBooks connection – [Loom](https://www.loom.com/share/ca1413e4c19247a08abb48e0133e381c?sid=f1596417-ff7c-4314-94b2-74ce3bf35797)


src/app/api/core/exceptions/api.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * Custom error class representing an error that occurred in Tasks API.
```
rrojan, line 2:
```suggestion
 * Custom error class representing an error that occurred in API.
```

src/app/api/quickbooks/auth/auth.controller.ts
```diff
@@ -0,0 +1,25 @@
+import authenticate from '@/app/api/core/utils/authenticate'
+import { AuthService } from '@/app/api/quickbooks/auth/auth.service'
+import httpStatus from 'http-status'
+import { NextRequest, NextResponse } from 'next/server'
+
+export async function getAuthorizationUrl(req: NextRequest) {
+  const user = await authenticate(req)
+  const token = req.nextUrl.searchParams.get('token') || ''
```
rrojan, line 8: Let's panic and throw 403 if `token` is not provided

src/app/api/quickbooks/auth/auth.controller.ts
```diff
@@ -0,0 +1,25 @@
...
+  const body = await req.json()
+  const redirectUrl = body?.redirectUrl || req.headers.get('referer')
+  const authService = new AuthService(user)
+  const authUrl = await authService.getAuthUrl({
+    token: token,
```
rrojan, line 13:
```suggestion
    token,
```
Optional

src/app/api/quickbooks/token/route.ts
```diff
@@ -0,0 +1,4 @@
+import { withErrorHandler } from '@/app/api/core/utils/withErrorHandler'
+import { checkPortalConnection } from './token.controller'
```
rrojan, line 2: Use aliased imports

src/app/api/quickbooks/token/token.controller.ts
```diff
@@ -0,0 +1,11 @@
+import authenticate from '@/app/api/core/utils/authenticate'
+import { TokenService } from '@/app/api/quickbooks/token/token.service'
+import { NextRequest, NextResponse } from 'next/server'
+import httpStatus from 'http-status'
+
+export async function checkPortalConnection(req: NextRequest) {
+  const user = await authenticate(req)
+  const tokenService = new TokenService(user)
+  const portalSync = await tokenService.checkPortalConnection(user.workspaceId)
+  return NextResponse.json(portalSync, { status: httpStatus.OK })
```
rrojan, line 10:
```suggestion
  return NextResponse.json(portalSync)
```
200 response is inferred by default,,. You can omit it unless this is a preferable style for you.

src/app/api/quickbooks/token/token.service.ts
```diff
@@ -0,0 +1,13 @@
...
+import { QBTokenSelectSchemaType } from '@/db/schema/qbTokens'
+import { getPortalConnection } from '@/db/service/token.service'
+
+export class TokenService extends BaseService {
+  async checkPortalConnection(
+    portalId: string,
+  ): Promise<QBTokenSelectSchemaType | null> {
+    const portalConnection = await getPortalConnection(portalId)
+
+    return portalConnection || null
```
rrojan, line 11: Is there a reason we are doing this additionally? `portalConnection` itself can be null right?

src/config/index.ts
```diff
@@ -1,3 +1,12 @@
...
-export const DATABASE_URL = process.env.DATABASE_URL
+export const appId = process.env.COPILOT_APP_API_KEY || ''
+export const databaseUrl = process.env.DATABASE_URL
+
+// Intuit credentials
+export const intuitClientId = process.env.INTUIT_CLIENT_ID || ''
+export const intuitClientSecret = process.env.INTUIT_CLIENT_SECRET || ''
+export const intuitRedirectUri = process.env.INTUIT_REDIRECT_URI || ''
+export const intuitEnvironment = process.env.INTUIT_ENVIRONMENT || ''
+export const intuitIntuitCompanyId = process.env.INTUIT_COMPANY_ID || ''
```
rrojan, line 12: Let's validate important env variables with zod so we get flagged when we forget to add it on deployed environments

src/db/index.ts
```diff
@@ -19,7 +23,7 @@ class DBClient {
     return DBClient.instance
   }
 
-  getDrizzle(): PostgresJsDatabase {
+  getDrizzle(): PostgresJsDatabase<typeof schema> {
```
rrojan, line 26: Let's abstract implementation details (optional)
```suggestion
  getDb / getInstance...
```

src/hook/useQuickbooks.ts
```diff
@@ -0,0 +1,77 @@
+import { copilotDashboardUrl } from '@/config'
+import { useState } from 'react'
+
+export const useQuickbooks = (token: string) => {
+  const [loading, setLoading] = useState<boolean>(false)
+  const [hasConnection, setHasConnection] = useState<boolean>(false)
```
rrojan, line 6: Boolean is inferred from the initial value so we don't really need to specify it verbosely. You can resolve this comment if you prefer to have it typed verbosely though.

src/hook/useQuickbooks.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  const getAuthUrl = async () => {
+    const redirectUrl = `${copilotDashboardUrl}`
```
rrojan, line 9:
```suggestion
    const redirectUrl = copilotDashboardUrl
```

src/utils/copilotAPI.ts
```diff
@@ -40,11 +40,8 @@ import { API_DOMAIN } from '@/constant/domains'
 export class CopilotAPI {
   copilot: SDK
 
-  constructor(
-    private token: string,
-    customApiKey?: string,
-  ) {
-    this.copilot = copilotApi({ apiKey: customApiKey ?? apiKey, token })
+  constructor(private token: string) {
+    this.copilot = copilotApi({ apiKey: apiKey, token })
```
rrojan, line 44:
```suggestion
    this.copilot = copilotApi({ apiKey, token })
```

rrojan (changes requested):

src/type/intuit.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'intuit-oauth'
```
rrojan, line 1: Let's add basic typing for the module as well

rrojan (changes requested):

src/app/api/quickbooks/token/token.service.ts
SandipBajracharya: Drizzle actually returns `undefined` if no records found. So need to send null if `undefined`.

src/config/index.ts
SandipBajracharya: It seems like we cannot directly parse using zod as the server env variable returns undefined by default and throws error in the UI. Need some work around for this. For now I've marked it as "TODO".

SandipBajracharya:
> "Go back to dashboard" button might be unnecessary right (unless its in the design / prd of course)... Wdyt?

That button is optional. It is in-case if user gets confused on where to go next. The design for the external redirection from QuickBooks is missing in the design. We can remove it later. Keeping it for now.

rrojan (approved):
LGTM 🏆

PR #4, opened 2025-04-28
Out 1655 | Handling edge-case scenarios and error

SandipBajracharya:
## Changes

- Handle access token expiry and refresh token
- Refresh token expiry and turn off sync status (change status to false)
- Supabase real time (Sync status change in real time after sync flag changes)
- Code refactor

## Testing Criteria

For the real time update after connection - [Loom](https://www.loom.com/share/f95e040cefb34c83bc5d81b6481c7889)


src/app/api/quickbooks/log/log.service.ts
```diff
@@ -0,0 +1,41 @@
...
+  ): Promise<QBConnectionLogSelectSchemaType> {
+    const [result] = await this.db
+      .update(QBConnectionLogs)
+      .set(log)
+      .where(
+        eq(
+          QBConnectionLogs.id,
+          sql`(SELECT ${QBConnectionLogs.id} From ${QBConnectionLogs} WHERE ${QBConnectionLogs.portalId} = ${log.portalId} 
+          AND ${QBConnectionLogs.connectionStatus} = ${ConnectionStatus.PENDING} ORDER BY ${QBConnectionLogs.createdAt} DESC LIMIT 1)`,
+        ),
```
rrojan, line 32: I might be misunderstanding here, but if there are no connection logs with PENDING status then the subquery will not return any rows, thus eq will always be false. We wouldn't be inserting in such a case right (the method name is upsert so I'm assuming it needs to insert in such a case)?

src/app/api/quickbooks/log/log.service.ts
```diff
@@ -0,0 +1,41 @@
...
+    log: QBConnectionLogCreateSchemaType,
+  ): Promise<QBConnectionLogSelectSchemaType> {
+    const [resultlog] = await this.db
+      .insert(QBConnectionLogs)
+      .values(log)
+      .returning()
+    return resultlog
+  }
+
+  async upsertConnectionLog(
```
rrojan, line 21:
```suggestion
  async upsertLatestConnectionLog(
```

src/app/api/quickbooks/log/route.ts
```diff
@@ -0,0 +1,4 @@
+import { withErrorHandler } from '@/app/api/core/utils/withErrorHandler'
+import { storeQbConnectionLog } from '@/app/api/quickbooks/log/log.controller'
+
+export const POST = withErrorHandler(storeQbConnectionLog)
```
rrojan, line 4: I couldn't find anywhere where this route is fetched... is this something we plan on doing later?

src/app/api/quickbooks/log/log.controller.ts
```diff
@@ -0,0 +1,13 @@
...
+import { QBConnectionLogCreateSchema } from '@/db/schema/qbConnectionLogs'
+import { NextRequest, NextResponse } from 'next/server'
+
+export const storeQbConnectionLog = async (req: NextRequest) => {
+  const user = await authenticate(req)
+  const body = await req.json()
+  const parsedBody = QBConnectionLogCreateSchema.parse(body)
+  const logService = new LogService(user)
+  const connectionLog = await logService.storeConnectionLog(parsedBody)
+  return NextResponse.json({ connectionLog })
```
rrojan, line 12: We can optionally return 201 here

rrojan (changes requested):

src/action/quickbooks.action.ts
```diff
@@ -1,15 +1,27 @@
...
+    console.error('checkPortalConnection#getPortalConnection | Error =', err)
     return null
   }
 }
+
+export async function checkSyncStatus(portalId: string): Promise<boolean> {
+  try {
+    const syncedPortal: QBTokenSelectSchemaType | null =
+      await getSyncedPortalConnection(portalId)
+    return !syncedPortal ? false : syncedPortal?.syncFlag || false
```
rrojan, line 22:
```suggestion
    const syncedPortal = await getSyncedPortalConnection(portalId)
    return syncedPortal?.syncFlag ?? false
```
Should be the same(?)

src/app/(home)/Home.tsx
```diff
@@ -10,26 +14,33 @@ export default async function Main({
...
+  const portalConnectionStatus =
     portalConnection && Object.keys(portalConnection).length > 0 ? true : false
 
+  const syncFlag = await checkSyncStatus(tokenPayload.workspaceId)
+
   return (
     <>
-      <HomeClient token={token} portalConnectionStatus={portalStatus} />
+      <HomeClient
+        token={token}
```
rrojan, line 39: Optionally in the future we can add this to a context using Context API so we can access this in any component without prop drilling too much

rrojan (changes requested):

src/app/api/quickbooks/log/log.service.ts
SandipBajracharya: In this function we are looking for those records with pending status for the portalId and if found, only update the latest. If not, insert new record. I will change the function name to something that matches its behavior to avoid the confusion.

src/app/api/quickbooks/log/route.ts
SandipBajracharya: I'd thought to store the log from client initially and created this API but changed later. Removing this.

src/app/api/quickbooks/log/log.controller.ts
SandipBajracharya: File removed.

src/app/api/quickbooks/log/log.service.ts
SandipBajracharya: Updated name to "upsertLatestPendingConnectionLog"

src/app/(home)/Home.tsx
SandipBajracharya: Yes! Thank you for the suggestion. Will implement this when working on more frontend related stuff.

rrojan (approved):
LGTM 🏆 
Amazing job

PR #8, opened 2025-05-08
OUT-1702 | OUT-1703 | Invoice webhook and payload (with line items)

SandipBajracharya:
## Changes

- [X] Invoice field mapping with line items
- [X] filter draft invoices
- [X] invoice mapping table update

### Invoice created in Copilot 
image

### invoice created in QuickBooks after webhook triggered 
image

### Mapped in table
image

### DIFF
--- a/src/app/api/core/types/client.ts
+++ b/src/app/api/core/types/client.ts
@@ -1,0 +1,4 @@
+export enum ClientStatus {
+  Active = 'active',
+  Invited = 'invited',
+}
--- a/src/app/api/core/types/invoice.ts
+++ b/src/app/api/core/types/invoice.ts
@@ -1,0 +1,6 @@
+export enum InvoiceStatus {
+  Draft = 'draft',
+  Open = 'open',
+  Paid = 'paid',
+  Void = 'void',
+}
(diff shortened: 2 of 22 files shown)


src/app/api/core/types/client.ts
```diff
@@ -0,0 +1,4 @@
+export enum ClientStatus {
+  Active = 'active',
+  Invited = 'invited',
+}
```
arpandhakal, line 4: I think keys of an enum should be in caps.

src/app/api/core/types/invoice.ts
```diff
@@ -0,0 +1,6 @@
+export enum InvoiceStatus {
+  Draft = 'draft',
+  Open = 'open',
+  Paid = 'paid',
+  Void = 'void',
+}
```
arpandhakal, line 6: I think keys of an enum should be in caps.

arpandhakal (approved):

src/app/api/core/types/client.ts
SandipBajracharya: You're right that it is a standard way to use the keys in uppercase but we can also use PascalCase.

PR #11, opened 2025-05-09
Out 1700 | Improvements

SandipBajracharya:
## Changes

- [X] Sentry setup.

## Testing Criteria

- Image defining Sentry reference error on undefined function:

image


sentry.server.config.ts
```diff
@@ -0,0 +1,23 @@
+// This file configures the initialization of Sentry on the server.
+// The config you add here will be used whenever the server handles a request.
+// https://docs.sentry.io/platforms/javascript/guides/nextjs/
+
+import * as Sentry from '@sentry/nextjs'
+
+const dsn = process.env.NEXT_PUBLIC_SENTRY_DSN || process.env.SENTRY_DSN
```
rrojan, line 7: Since DSN isn't a sensitive name, let's just keep one env var with `NEXT_PUBLIC`, unless we're planning to use separate projects for client & server

rrojan (changes requested):

rrojan:
Let's also add the sentry env vars to prod & preview environments so we can test them there

sentry.server.config.ts
SandipBajracharya: Removed

rrojan (approved):
LGTM 🏆

rrojan:
Let's use conventional commits in this format so we can more easily pinpoint breaking changes, etc in the future

`feat|fix|perf|docs|test|chore|refactor|hotfix:(TIC-001): some change`

PR #12, opened 2025-05-09
OUT-1704 | Include client information in Invoice payload

SandipBajracharya:
## Changes

- [X] Check if customer exist using custom query, against given name and family name
- [X] Create customer in QB if does not exist

## Testing Criteria

- [X] Case when customer does not exist in QB. (Creates new one) [Loom](https://www.loom.com/share/67881974baaa4cdf95fb3aa5e4cbca15?sid=39779b39-6db9-4d03-8894-acfe7456dee4)
- [X] Case when customer exists. [Loom](https://www.loom.com/share/da725a01589240b69039e542bde05690?sid=b93e9b6b-3971-4a93-9122-a27bf460bd58)

## Notes
The recipient name here is actually different because Invoice creation is triggered from staging environment and we need production tokens for copilot SDK to work. So fetching client info from production environment makes customer name different


rrojan (approved):
LGTM 🏆

PR #13, opened 2025-05-12
OUT-1748 | Customer update in QB if change in copilot side

SandipBajracharya:
## Changes

- [X] mapping table (qb_customers)
- [X] update customer info (familyName, givenName, email) in QB if change in copilot side

## Testing Criteria

- mapping table 
image

- update user with old full name in detail
[Loom](https://www.loom.com/share/d288ef71b5354fdeb977a7fd0bee0bbd?sid=73e55d4b-97f9-4ba4-b590-7e89b70ebb14)

### DIFF
--- a/src/app/api/quickbooks/customer/customer.service.ts
+++ b/src/app/api/quickbooks/customer/customer.service.ts
@@ -38,10 +38,7 @@ export class CustomerService extends BaseService {
     conditions: WhereClause,
     returningFields?: (keyof typeof QBCustomers)[],
   ) {
-    console.log(
-      'CustomerService#updateQBCustomer | For client with Id =',
-      payload?.clientId,
-    )
+    console.log('CustomerService#updateQBCustomer')
     const parsedInsertPayload = QBCustomerUpdateSchema.parse(payload)
 
     const query = this.db
--- a/src/app/api/quickbooks/customer/customer.service.ts
+++ b/src/app/api/quickbooks/customer/customer.service.ts
@@ -1,0 +1,77 @@
+import { BaseService } from '@/app/api/core/services/base.service'
+import { buildReturningFields } from '@/db/helper/drizzle.helper'
+import {
+  QBCustomerCreateSchema,
+  QBCustomerCreateSchemaType,
+  QBCustomers,
+  QBCustomerUpdateSchema,
+  QBCustomerUpdateSchemaType,
+} from '@/db/schema/qbCustomers'
+import { and, isNull, SQL } from 'drizzle-orm'
+
+type WhereClause = SQL<unknown>
+
+export class CustomerService extends BaseService {
+  async createQBCustomer(
+    payload: QBCustomerCreateSchemaType,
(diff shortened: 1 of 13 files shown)


src/app/api/quickbooks/invoice/invoice.service.ts
```diff
@@ -89,42 +92,122 @@ export class InvoiceService extends BaseService {
...
+          PrimaryEmailAddr: {
+            Address: client.email,
+          },
         }
+
+        const customerRes = await intuitApi.createCustomer(customerPayload)
+        customer = customerRes.Customer
+
+        if (!customer) {
+          throw new Error('Failed to create customer')
```
rrojan, line 144: This can be our custom API Error instead

src/app/api/quickbooks/invoice/invoice.service.ts
```diff
@@ -89,42 +92,122 @@ export class InvoiceService extends BaseService {
...
+      if (Object.keys(sparseUpdatePayload).length > 0) {
+        const customerSparsePayload = {
+          ...sparseUpdatePayload,
+          Id: existingCustomer.qbCustomerId,
+          SyncToken: existingCustomer.qbSyncToken,
+          DisplayName: `${client.givenName} ${client.familyName}`,
+          BillAddr: {
+            Line1: `${existingCustomer.givenName} ${existingCustomer.familyName}`,
+            Line2: existingCustomer.companyName,
+          },
```
rrojan, line 187: Maybe we should consider adding more address details from `Line3` onwards as well?
image

src/app/api/quickbooks/customer/customer.service.ts
```diff
@@ -0,0 +1,74 @@
...
+import {
+  QBCustomerCreateSchema,
+  QBCustomerCreateSchemaType,
+  QBCustomers,
+  QBCustomerUpdateSchema,
+  QBCustomerUpdateSchemaType,
+} from '@/db/schema/qbCustomers'
+import { and, isNull, SQL } from 'drizzle-orm'
+
+type WhereClause = SQL<unknown>
```
rrojan, line 12: Is there a reason we are abstracting SQL in this way?

rrojan (changes requested):

src/app/api/quickbooks/customer/customer.service.ts
SandipBajracharya: I'm passing the where condition into a function that has type SQL<unknown>. So did this just to match the params type. I think only SQL will also support the SQL<unknown> type param. I'll make sure on this one. Thanks for pointing out.

image

src/app/api/quickbooks/invoice/invoice.service.ts
SandipBajracharya: Actually there is no mention of address info in PRD so have not included that for now. But adding that address info is good idea. 
image

rrojan (approved):
LGTM 🏆

PR #16, opened 2025-05-20
Out 1753 | Flatten copilot product api with price

SandipBajracharya:
## Changes

- [X] Prepare an api with flatten product and price
- [X] use of Upstash/Redis for caching


src/app/api/quickbooks/product/product.controller.ts
```diff
@@ -0,0 +1,16 @@
...
+import { ProductService } from '@/app/api/quickbooks/product/product.service'
+import { NextRequest, NextResponse } from 'next/server'
+
+export async function getFlattenProducts(req: NextRequest) {
+  const user = await authenticate(req)
+  const productService = new ProductService(user)
+  const nextToken = req.nextUrl.searchParams.get('nextToken') || undefined
+  const limit = req.nextUrl.searchParams.get('limit') || MAX_PRODUCT_LIST_LIMIT
+  const products = await productService.getFlattenProductList(
+    Number(limit),
```
rrojan, line 12: Very niche issue, but if we send a non-numeric string in limit instead, `Number(limit)` would be `NaN`, which would be passed to `getFlattenProductList`

src/app/api/quickbooks/product/product.controller.ts
```diff
@@ -0,0 +1,16 @@
+import { MAX_PRODUCT_LIST_LIMIT } from '@/app/api/core/constants/limit'
+import authenticate from '@/app/api/core/utils/authenticate'
+import { ProductService } from '@/app/api/quickbooks/product/product.service'
+import { NextRequest, NextResponse } from 'next/server'
+
+export async function getFlattenProducts(req: NextRequest) {
+  const user = await authenticate(req)
+  const productService = new ProductService(user)
+  const nextToken = req.nextUrl.searchParams.get('nextToken') || undefined
+  const limit = req.nextUrl.searchParams.get('limit') || MAX_PRODUCT_LIST_LIMIT
```
rrojan, line 10:
```suggestion
  const searchParams = req.nextUrl.searchParams
  const nextToken = searchParams.get('nextToken') || undefined
  const limit = searchParams.get('limit') || MAX_PRODUCT_LIST_LIMIT
```
Optional change

src/utils/copilotAPI.ts
```diff
@@ -315,6 +319,23 @@ export class CopilotAPI {
...
+    nextToken?: string,
+    limit?: number,
+  ): Promise<ProductsResponse | undefined> {
+    try {
+      console.info('CopilotAPI#getProducts | token =', this.token)
+      return ProductsResponseSchema.parse(
+        await this.copilot.listProducts({ name, nextToken, limit }),
+      )
+    } catch (error: unknown) {
+      const tError = error as CopilotApiError
```
rrojan, line 333: Let's use `instanceOf` or `typeof error.X === 'Y'` approach instead. There are chances this might be other forms of errors

src/utils/copilotAPI.ts
```diff
@@ -315,6 +319,23 @@ export class CopilotAPI {
...
+      console.error('CopilotAPI#getProducts | message =', tError.body.message)
+      return
```
rrojan, line 335: Is there a reason why we are just console.error'ing an error and returning undefined?

src/app/api/quickbooks/product/product.service.ts
```diff
@@ -72,4 +74,38 @@ export class ProductService extends BaseService {
...
+
+  async getFlattenProductList(limit: number, nextToken?: string) {
+    // 1. check in redis. key (${workspaceId}_productList) if exists, return the value
+    const redisKey = `${this.user.workspaceId}-productList`
+    const redisClient = RedisClient.getInstance()
+
+    const cachedProducts = await redisClient.get(redisKey)
+    if (cachedProducts) {
+      return cachedProducts // auto deserialization by default
+    }
```
rrojan, line 86: Might want to validate the data here to ensure type safety... i think the default return of `redisClient.get` would be `any` here

src/app/api/quickbooks/product/product.service.ts
```diff
@@ -72,4 +74,38 @@ export class ProductService extends BaseService {
...
+
+    // 2. if not exists, get all the products from copilot and store in redis
+    const copilot = new CopilotAPI(this.user.token)
+    const products = await copilot.getProducts(undefined, nextToken, limit)
+    const flattenProductsPrice = (
+      await Promise.all(
+        (products?.data ?? []).map(async (product) => {
+          const prices = await copilot.getPrices(product.id)
```
rrojan, line 94: We will hit Copilot rate limits in `Promise.all` if the number of products are high (20+)

rrojan (changes requested):

src/app/api/quickbooks/product/product.service.ts
rrojan: I recommend using the bottleneck package while hitting Copilot API in a Promise.all. The ratelimit is supposed to be 20req/s, and even this should be lowered in a prod environment since there are multiple users that could be using our app with the same API key in prod

https://www.npmjs.com/package/bottleneck

src/utils/copilotAPI.ts
SandipBajracharya: You're right. I've modified the code now with safe attribute access.

src/utils/copilotAPI.ts
SandipBajracharya: returning undefined cause there are few process that we don't want to terminate if there are status like 400 or 404. There should be console.info in this case. I will modify this. Thanks

src/app/api/quickbooks/product/product.service.ts
SandipBajracharya: Removed redis.

PR #21, opened 2025-05-23
Out 1792 | Create product in QB when created in copilot

SandipBajracharya:
## Changes

- [X] Listen to product.created and price.created webhook and create product in qb
- [X] Map the product in mapping table

## Testing Criteria

- [X] Item created in QB when created in copilot [Loom](https://www.loom.com/share/b8b950c494c946af81f416e10bd4e1cb?sid=557f1a89-ba52-4b7f-8d86-4d51497cbc6e)

## Notes
- To create product, we must also listen to price.created web hook as product creation doesn't include the price info


src/app/api/quickbooks/product/product.service.ts
```diff
@@ -236,8 +245,8 @@ export class ProductService extends BaseService {
           }
 
           const fullUpdatePayload: QBItemFullUpdatePayloadType = {
-            Id: product.qbItemId,
-            SyncToken: product.qbSyncToken,
+            Id: product.qbItemId as string,
+            SyncToken: product.qbSyncToken as string,
```
rrojan, line 249: Would be better to parse as string with zod

PR #25, opened 2025-05-30
Out 1800 | Out 1801: Integrate product listing APIs

SandipBajracharya:
## Changes

- [X] show the list of copilot products in the left side of the product mapping table
- [X] show the list of QB items in drop down option for user to map with the copilot product.

## Testing Criteria

[Loom](https://www.loom.com/share/98d8f540530c45f48c17d00b06a26697?sid=cbeae0c4-4778-49b1-aa6e-21dc01491cba)


src/components/dashboard/settings/sections/product/ProductMappingTable.tsx
```diff
@@ -104,8 +69,11 @@ export default function ProductMappingTable() {
 
   const getFilteredItems = (index: number) => {
     const searchTerm = searchTerms[index] || ''
-    return quickbooksItems.filter((item) =>
-      item.name.toLowerCase().includes(searchTerm.toLowerCase()),
+    return (
+      quickbooksItems &&
+      quickbooksItems.filter((item) =>
+        item.name.toLowerCase().includes(searchTerm.toLowerCase()),
```
rrojan, line 75: Let's trim these strings as well, especially the searchTerm

src/components/dashboard/settings/sections/product/ProductMappingTable.tsx
```diff
@@ -133,102 +101,125 @@ export default function ProductMappingTable() {
...
-                        <div className="text-sm">
-                          {selectedItems[index].name}
-                        </div>
-                        <div className="text-sm text-gray-500">
-                          {selectedItems[index].price}
-                        </div>
+              </tr>
+            ) : products ? (
+              products.map((product, index) => (
+                <tr key={index} className="hover:bg-gray-50 transition-colors">
```
rrojan, line 114: These ternary statements are going too deep... can we look into simplifying it

src/components/dashboard/settings/sections/product/ProductMappingTable.tsx
```diff
@@ -133,102 +101,125 @@ export default function ProductMappingTable() {
...
+                  </td>
+                </tr>
+              ))
+            ) : (
+              <tr className="text-center">
+                <td colSpan={3} className="py-11">
+                  Start by creating a product in Copilot.
+                  <a href="#" className="ms-2">
+                    Create Product
+                  </a>
```
rrojan, line 219: We can implement the app bridge for linking from iframe to a Copilot route right now, or create a followup ticket for it

https://docs.copilot.app/docs/custom-apps-developing-your-first-app#app-bridge

However, I could only see an app bridge linking to billing section, not specifically to the products page... might have to investigate further on this

src/helper/swr.helper.ts
```diff
@@ -0,0 +1,9 @@
+import { getFetcher } from '@/helper/fetch.helper'
+import useSWR from 'swr'
+
+const fetcher = (url: string) => getFetcher(url, {})
+
+export const useSwrHelper = (key: any) =>
+  useSWR(key, fetcher, {
+    revalidateOnFocus: false,
```
rrojan, line 8: Personally I don't see a good enough use case for this wrapper for useSWR... this doesn't support additional props to be passed to the useSWR options as well... we should either use the hook directly or accept an optional `opts` arg here that will be destructured after `revalidateOnFocus` imo

rrojan (changes requested):

src/helper/swr.helper.ts
SandipBajracharya: Great opinion. I was thinking of the same: to pass the opts param to make the most out of the helper function. The reason I created this was to avoid using options like `revalidateOnFocus` repeatedly.

src/components/dashboard/settings/sections/product/ProductMappingTable.tsx
SandipBajracharya: The ternary is only two level deep. There are other ternary operators inside the block though if that is your concern?

rrojan (approved):
LGTM 🏆

PR #27, opened 2025-06-02
Out 1803 | Product exclude feature from mapping and invoice

SandipBajracharya:
## Changes

- [X] add "is_excluded" flag in product sync table
- [X] mark excluded column flag to true and do not create/ map item in QB when invoice created. The line items should follow default behavior (i.e. as OneOffItem)

## Testing Criteria

[Loom](https://www.loom.com/share/418d9eb4c4e64d0698b1939c748a64bb)

### DIFF
--- a/src/app/api/quickbooks/auth/auth.service.ts
+++ b/src/app/api/quickbooks/auth/auth.service.ts
@@ -84,6 +84,9 @@ export class AuthService extends BaseService {
         // query income acc ref from intuit
         const incomeAccRef = await intuitApi.getSingleIncomeAccount()
         insertPayload.incomeAccountRef = incomeAccRef.Id
+      } else {
+        // case for re-authorization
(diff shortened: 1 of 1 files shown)


src/hook/useSettings.ts
```diff
@@ -162,6 +180,21 @@ export const useProductTableSetting = () => {
...
+      return {
+        name: null,
+        priceId: product.priceId,
+        productId: product.id,
+        unitPrice: null,
+        qbItemId: null,
+        qbSyncToken: null,
+        isExcluded: true,
+      }
+    })
```
arpandhakal, line 194: instead of parsing into an 'any' type. Can we use a zod schema to parse each items in products to a certain schema? This promotes type safety, catches bad data at runtime and avoid unknown errors.

arpandhakal (approved):

PR #28, opened 2025-06-04
feat(OUT-1807) : track syc since when it has been active

arpandhakal:
## Changes

- [x] added an api to fetch latest success log for a connected portal which returns `createdAt` of the latest `successLog` in response.
- [x] fetched the latest success log in the main server component.
- [x] managed `useAuth` and `useQuickbooks` hooks to manage the state of the current `successLog` in realTime.
- [x] used `react-time-ago` to show the date string in (present time - createdAt) time ago in the success callout component.

## Testing Criteria

image

- WIP : Realtime testing


src/app/(home)/Home.tsx
```diff
@@ -7,14 +7,23 @@ import {
...
+  return (await response.json()).data
+}
+
 export default async function Main({
   searchParams,
 }: {
   searchParams: Promise<{ token: string; type?: string }>
 }) {
   const { token, type } = await searchParams
+  const successLog = await getLatestSuccesLog(token)
```
SandipBajracharya, line 26: I think we can move this after sync check. Success log doesn't make sense if the sync is off. There will be one less api call if sync is off. Check line no. 55 of the same file

src/app/(home)/Home.tsx
```diff
@@ -56,6 +64,7 @@ export default async function Main({
         syncFlag={syncFlag}
         reconnect={reconnect}
         portalConnectionStatus={portalConnectionStatus}
+        lastSyncTimestamp={successLog?.createdAt || null}
```
SandipBajracharya, line 67: After implementing the previous comment, we might not need to check for undefined successLog if the sync is on. This is for safety so change is optional.

src/hook/useQuickbooks.ts
```diff
@@ -53,6 +56,10 @@ export const useQuickbooks = (
               : newPayload.connection_status === ConnectionStatus.PENDING
                 ? false
                 : null
+          const lastSuccessTimestamp = connectionStatus
+            ? newPayload.createdAt
+            : null
+          setLastSyncTimestampState(lastSuccessTimestamp) //realtime update of last sync timestamp
```
SandipBajracharya, line 62: Since we are using the `lastSyncTimestamp` param in Auth context, I think we can utilize that state. We can simply `lastSyncTimestamp` value from `setAuthParams()` setter and use it in `useDashboard.ts` hook

SandipBajracharya:
Awesome work :trophy: 

IMO, we can directly use and fetch from the server component without creating an api. This will also be equally safe as we have done token parse and check in Home.tsx. But this is totally optional. We might even need the api later for other reasons.

Also the last sync date is not reflected when the connection is success. However, it does show in initial load. Lets implement the changes and we can even do peer testing on this.

src/app/(home)/Home.tsx
```diff
@@ -48,6 +57,8 @@ export default async function Main({
     }
```
SandipBajracharya, line 57: What if we do something like below? This will prevent from api call if sync turn off

```
let successLog
if (!syncFlag) {
  ...
} else {
  successLog = await getLatestSuccesLog(token)
}
```
And for AuthProvider (same as before):

```
lastSyncTimestamp={successLog?.createdAt || null}
```

PR #40, opened 2025-06-12
OUT-1865 | "Confirm" button should disappear after click

SandipBajracharya:
## Changes

- [X] Once mapping is confirmed, hide "Confirm" button
- [X] Allow "Confirm" even if no items mapped for the first mapping after connection
- [X] Hide/unhide "Confirm" as per the change in mapping table

## Testing Criteria

[Loom] (https://www.loom.com/share/67f55750c98e48b4a803806d856a8d99?sid=807d2641-a4b0-4091-9e45-2fd75ba3e894)


src/app/context/AuthContext.tsx
```diff
@@ -10,6 +11,8 @@ type AuthContextType = {
   portalConnectionStatus: boolean
   isEnabled: boolean
   lastSyncTimestamp: string | null
+  showProductConfirm?: boolean
+  initialProductMap?: ProductMappingItemType[]
```
rrojan, line 15: This works but calling this AuthContext might be a bit misleading since these are not related to auth. Let's do a quick LSP rename for it to something that suits better like AppContext?

PR #44, opened 2025-06-18
Out 1740 | Create sync log table

SandipBajracharya:
## Changes

- [X] Create sync log table

## Testing Criteria

- Table creation:

image


src/app/api/quickbooks/invoice/invoice.service.ts
```diff
@@ -532,19 +532,21 @@ export class InvoiceService extends BaseService {
...
-      },
-      eq(QBInvoiceSync.id, invoiceSync.id),
-      ['id'],
-    )
+    await Promise.all([
+      paymentService.createPaymentAndSync(
+        intuitApi,
+        qbPaymentPayload,
+        payload.data.number,
+      ),
```
rrojan, line 541: Is there a chance that the payment could still fail here? In that case we wouldn't want to update the QB invoice with the status of PAID.

SandipBajracharya: Yes you're right. In my PR to re-sync, I've actually fixed this. Thanks for pointing out!

PR #49, opened 2025-06-23
OUT-1860 | Sync invoices assigned to company

SandipBajracharya:
## Changes

- [X] Create customer in QB if invoice is assigned to company if "use company name" setting is enabled
- [X] If "use company name" flag is disabled, get the first client of the company and use clients detail
- [X] Sparse update the customer info in QuickBooks if customer is already created and "use company name" setting value changes

## Testing Criteria

WIP


src/app/api/quickbooks/invoice/invoice.service.ts
```diff
@@ -267,71 +287,112 @@ export class InvoiceService extends BaseService {
...
 
       // 3. if not found, create a new client in the QB
       if (!customer) {
         if (!company) {
           // Case when client is retrieved directly from recipientId
-          company = await this.copilot.getCompany(client.companyId)
+          company = await this.copilot.getCompany(clientCompany.companyId)
 
           if (!company) {
             // Indicates company is not available in any case. This only logs as error and allows to create customer in QB since company name is optional
```
rrojan, line 361:
```suggestion
          if (!company || !company.name) {
            // Indicates company is not available in any case. This only logs as error and allows to create customer in QB since company name is optional
```
This is a weird edge case with Copilot.

Clients without companies are also assigned `companyId`, and upon doing `copilot.getCompany`, it returns a valid company payload for many clients, but with the `name` as an empty string.

Similar implementation in Tasks:

image

src/app/api/quickbooks/invoice/invoice.service.ts
```diff
@@ -345,23 +406,23 @@ export class InvoiceService extends BaseService {
         'Id' | 'SyncToken' | 'sparse'
       > = {}
 
-      if (existingCustomer.familyName !== client.familyName) {
-        sparseUpdatePayload.FamilyName = client.familyName
+      if (existingCustomer.familyName !== clientCompany.familyName) {
+        sparseUpdatePayload.FamilyName = clientCompany.familyName
```
rrojan, line 410:
```suggestion
        sparseUpdatePayload.FamilyName = clientCompany.familyName
```
This is basically the same thing but simpler

src/app/api/quickbooks/invoice/invoice.service.ts
```diff
@@ -517,15 +577,15 @@ export class InvoiceService extends BaseService {
     }
 
     // 2. if not, create payment in QB, sync payment in payment sync table and change invoice status to paid
-    if (!invoiceSync?.clientId) {
+    if (!invoiceSync?.recipientId) {
```
rrojan, line 580:
```suggestion
    if (!invoiceSync.recipientId) {
```
invoiceSync is guaranteed to be present, since we have a check for it above

rrojan (changes requested):

rrojan (approved):
Please merge after fixes

PR #51, opened 2025-06-25
OUT-1836 | Implementation for deleted invoices

rrojan:
## Changes

- [x] Support invoice deletion webhook and sync deletion to QB

## Testing Criteria

- [x] Screencast

https://github.com/user-attachments/assets/9c36f131-a8bd-420c-819d-8dbea79b5217

### DIFF
--- a/src/app/api/quickbooks/invoice/invoice.service.ts
+++ b/src/app/api/quickbooks/invoice/invoice.service.ts
@@ -21,8 +21,8 @@ import {
 import { TransactionType, WhereClause } from '@/type/common'
 import {
   QBCustomerSparseUpdatePayloadType,
+  QBDestructiveInvoicePayloadSchema,
   QBNameValueSchemaType,
-  QBVoidInvoicePayloadSchema,
 } from '@/type/dto/intuitAPI.dto'
 import {
   InvoiceCreatedResponseType,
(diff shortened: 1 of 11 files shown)


src/app/api/quickbooks/invoice/invoice.service.ts
```diff
@@ -644,4 +645,57 @@ export class InvoiceService extends BaseService {
...
+
+    if (!safeParsedPayload.success) {
+      throw new APIError(
+        httpStatus.INTERNAL_SERVER_ERROR,
+        'WebhookService#handleInvoiceDeleted | Could not parse invoice delete payload',
+      )
+    }
+
+    await Promise.all([
+      intuitApi.deleteInvoice(safeParsedPayload.data),
```
SandipBajracharya, line 691: If invoice delete fails in QB, we don't want to update our invoice sync table. I think we can remove parallel processing here.

src/app/api/quickbooks/webhook/webhook.service.ts
```diff
@@ -374,4 +380,23 @@ export class WebhookService extends BaseService {
         console.error('WebhookService#handleWebhookEvent | Unknown event type')
     }
   }
+
+  private async handleInvoiceDeleted(
+    payload: unknown,
```
SandipBajracharya, line 385: Even if there is safeParse() used, I think we can still use some proper type here instead of 'unknown'. wdyt?

src/type/dto/intuitAPI.dto.ts
```diff
@@ -124,6 +124,15 @@ export type QBVoidInvoicePayloadType = z.infer<
   typeof QBVoidInvoicePayloadSchema
 >
 
+export const QBDeleteInvoicePayloadSchema = z.object({
```
SandipBajracharya, line 127: These types have not been used in code. Can you please check?

src/app/api/quickbooks/webhook/webhook.service.ts
rrojan: Given that we parse the payload in the function itself, and the type of the payload is not validated until then, I think it's okay to keep it as unknown here, especially since the function is not intended to be reusable

image

SandipBajracharya (approved):

PR #56, opened 2025-06-26
Out 1925 | Manage and regenerate migration files and rename tables

SandipBajracharya:
## Changes

- [X] Regenerate migration files with timestamp prefix
- [X] Also rename table "qb_tokens". This table does not only hold the token info but also QB account ids like expense account id, income account id.

## Testing Criteria

WIP


src/db/schema/qbConnectionLogs.ts
```diff
@@ -18,7 +18,7 @@ export const connectionStatusEnum = t.pgEnum(
 )
 
 export const QBConnectionLogs = table('qb_connection_logs', {
-  id: t.integer().primaryKey().generatedAlwaysAsIdentity(),
+  id: t.uuid().defaultRandom().primaryKey(),
```
rrojan, line 21: Does this mean we have to remove existing data from all or db + other devs' environments?

rrojan (approved):
LGTM 🏆

PR #60, opened 2025-07-03
OUT-1980: Invoice assigned to company are not syncing to QB

SandipBajracharya:
## Changes

- when company name setting flag is false and invoice billed to company then customer will be create in QB as follows: customer name (company name)
- if the company name setting flag is true, customer is created/updated using company name

## Testing Criteria

image (1)


rrojan (approved):
LGTM!

PR #67, opened 2025-07-09
OUT-1974: improve syncing with Quickbooks feedback

SandipBajracharya:
## Changes

- [X] added loading text when loading time is more than 3 seconds

## Testing Criteria

[Loom](https://www.loom.com/share/546208a1c3bc4a4c9a686b0f60dc5f4c?sid=c1de383e-c381-4fe0-8ca3-cd1535110a05)

### DIFF
--- a/src/hook/useSettings.ts
+++ b/src/hook/useSettings.ts
@@ -283,6 +283,7 @@ export const useProductMappingSettings = () => {
 export const useProductTableSetting = (
   setMappingItems: (mapProducts: ProductMappingItemType[]) => void,
 ) => {
+  const [showLoadingText, setShowLoadingText] = useState<boolean>(false)
   const emptyMappedItem = {
     name: null,
     description: '',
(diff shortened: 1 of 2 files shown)


src/hook/useSettings.ts
```diff
@@ -376,6 +377,16 @@ export const useProductTableSetting = (
     }
   }, [products, mappedItems])
 
+  useEffect(() => {
+    const timeout = setTimeout(() => {
+      setShowLoadingText(true)
+    }, 3000) // As general rule, if loading takes more than 3 seconds, show loading text
```
sajxraj, line 383: isn't 3 seconds to much? i don't know the context, wont it be better to show loading right away?

SandipBajracharya: The loading spinner is shown right away. If the loading takes more than 3 seconds, the loading text is expected as shown in the loom. Showing loading text if loading time is less than 3 seconds won't do any harm but might not be necessary. Optimization is necessary here to reduce the loading time, but that will be covered in the future milestones.

PR #68, opened 2025-07-10
OUT-1958: Hide the action menu when the app is disconnected

SandipBajracharya:
## Changes

- [X] When the app is not connected, hide all action menus.
- [X] When the app is connected but not enabled, hide the "Disconnect app" menu.

## Testing Criteria

- When the app is not connected
image

- When the app is not enabled
image

- When the app is connected and enabled
image


rrojan (commented):
LGTM 🏆

PR #71, opened 2025-07-15
OUT-2025: QB item's price is not mapping for the product

SandipBajracharya:
## Changes

- [X] use mapped items description, name and price when invoice generate
- [X] tax based on the mapped items (using tax percentage)
- [X] fixed mapped item price format

## Testing Criteria

[Loom](https://www.loom.com/share/8f3d99b358f345d49228337fcdf25442?sid=1bee96d1-0de0-4f1b-8eed-3b861ca43d1e)

- Price format fix
image

### DIFF
--- a/src/type/dto/webhook.dto.ts
+++ b/src/type/dto/webhook.dto.ts
@@ -33,6 +33,7 @@ export const InvoiceCreatedResponseSchema = z.object({
     status: z.nativeEnum(InvoiceStatus),
     total: z.number(),
     taxAmount: z.number().optional(),
+    taxPercentage: z.number().optional(),
     sentDate: z.string().datetime().nullish(),
     dueDate: z.string().datetime().nullish(),
   }),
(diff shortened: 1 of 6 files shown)


src/type/dto/webhook.dto.ts
```diff
@@ -33,6 +33,7 @@ export const InvoiceCreatedResponseSchema = z.object({
     status: z.nativeEnum(InvoiceStatus),
     total: z.number(),
     taxAmount: z.number().optional(),
+    taxPercentage: z.number().optional(),
```
rrojan, line 36:
```suggestion
    taxPercentage: z.number().default(0),
```

Up to preference, but this will save us from having to check if `taxPercentage` is undefined and running ternary checks wherever this field is used.

In further calculation, anyNumber * 0 will be 0

rrojan (approved):

PR #81, opened 2025-07-23
OUT-2056: Initial "Connect to Quickbooks" screen

SandipBajracharya:
## Changes

- [X] update initial connection screen with settings section and product mapping table
- [X] blurred and disabled setting section
- [X] issue fix: when sync fails, sync flag was not true after re-authorization
- [X] issue fix: when sync fails and re-authorization is done, the QB items in dropdown were not loaded.
- [X] only fetch QB items when sync flag is true. Added condition to not fetch QB items when:
	- sync fails
	- during initial page
- [X] "Connecting" label in callout button when initial connection and re-authorization

## Testing Criteria

[Loom](https://www.loom.com/share/a47a95886b034cfea5ceec336bd68a3c?sid=c6e87d7c-96f5-4bec-83f7-5627d8f28d87)

### DIFF
--- a/src/app/(home)/HomeClient.tsx
+++ b/src/app/(home)/HomeClient.tsx
@@ -1,25 +1,10 @@
 'use client'
 import { useApp } from '@/app/context/AppContext'
 import { Main as DashboardMain } from '@/components/dashboard/Main'
-import { SilentError } from '@/components/template/SilentError'
-import { useAppBridge, useQuickbooks } from '@/hook/useQuickbooks'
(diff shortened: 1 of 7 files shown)


rrojan (commented):
https://github.com/user-attachments/assets/0efeac23-9ab3-455e-86cf-02969ad8bda5

Since there is no sync with quickbooks going on when sync is not active, maybe we should consider changing the copy here to "Fetching Copilot Products" instead? Out of the scope of current ticket though.

rrojan (approved):
LGTM 🏆 
Great PR!

PR #88, opened 2025-07-30
OUT-2088: Optimistically hide update sync button

SandipBajracharya:
## Changes

- [X] Optimistically hide update sync button

## Testing Criteria

[Loom](https://www.loom.com/share/c68430a93c104186a5cbca6dc2c2d206?sid=50d79173-dabd-4774-bf6b-fe6b7604be1e)

### DIFF
--- a/src/hook/useSettings.ts
+++ b/src/hook/useSettings.ts
@@ -146,8 +146,11 @@ export const useProductMappingSettings = () => {
       )
       setChangedItemReference([])
     } else {
-      console.error({ tableRes, settingRes })
-      alert('Error submitting mapping items')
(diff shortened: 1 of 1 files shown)


src/hook/useSettings.ts
```diff
@@ -129,6 +129,11 @@ export const useProductMappingSettings = () => {
   }
 
   const submitMappingItems = async () => {
```
arpandhakal, line 131: Should we handle roll back functionality in case of an error. lets say the optimistic update has already hidden the update sync button and the update sync functionality fails. Should we revert this in that specific scenario?

PR #91, opened 2025-07-31
OUT-2108: Potential fix for products are occassionally not being created

SandipBajracharya:
## Changes

- [X] only listen to price.created webhook. Removed codes related product.created. Product info fetched from copilot by productId received from price payload.


src/app/api/quickbooks/product/product.service.ts
```diff
@@ -73,10 +74,25 @@ export class ProductService extends BaseService {
 
     return await this.db.query.QBProductSync.findFirst({
       where: conditions,
+      orderBy: [desc(QBProductSync.createdAt)],
       ...columns,
     })
   }
 
+  async getOneByProductId(
```
arpandhakal, line 82: This isn't being used anywhere. Is this for future use?

PR #113, opened 2025-08-07
OUT-2129: Cron job fails to re-sync failed records and code refactor

SandipBajracharya:
## Changes

- [X] followed Single responsibility priniciple for the functions for code refactor
- [X] proudct re-sync issue fix when cron runs. Since we are processing product created with price.created webhook, the mechanisms to re-sync also needed some change

## Testing Criteria

wip

### DIFF
--- a/src/app/api/quickbooks/cron/cron.service.ts
+++ b/src/app/api/quickbooks/cron/cron.service.ts
@@ -13,10 +13,12 @@ export default class CronService {
       workspaceId,
     }
     const token = encodePayload(copilotAPIKey, payload)
+    console.info({ token })
 
     // check if token is valid or not
     const copilot = new CopilotAPI(token)
     const tokenPayload = await copilot.getTokenPayload()
+    console.info({ tokenPayload })
     if (!tokenPayload) throw new APIError(500, 'Encoded token is not valid') // this should trigger p-retry and re-run the function
 
     const user = new User(token, tokenPayload)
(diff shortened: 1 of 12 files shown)


src/app/api/quickbooks/cron/cron.service.ts
```diff
@@ -13,10 +13,12 @@ export default class CronService {
       workspaceId,
     }
     const token = encodePayload(copilotAPIKey, payload)
+    console.info({ token })
```
arpandhakal, line 16: Some identifier in these console infos like "COPILOT_API_CRON_TOKEN" would be good to help us identify these in logs easily

src/utils/copilotAPI.ts
```diff
@@ -421,16 +421,9 @@ export class CopilotAPI {
...
   ): Promise<InvoiceResponse[] | undefined> {
     console.info('CopilotAPI#getInvoices | token =', this.token)
-    const data = await this.manualFetch(
-      'invoices',
-      {
-        limit: MAX_INVOICE_LIST_LIMIT.toString(),
-      },
-      workspaceId,
-    )
+    const data = await this.manualFetch('invoices', undefined, workspaceId)
```
arpandhakal, line 424: Is removing the limit query intended here? Also, I see there is a built in method of retrieving invoice, listing invoice in copilot node sdk. Is manual fetch necessary here?

arpandhakal (approved):

src/utils/copilotAPI.ts
SandipBajracharya: Copilot sdk actually doesn't support limit param for invoice list, that's why I've used the manual fetch. Also readding limit param. I was testing without limit.

src/app/api/quickbooks/webhook/webhook.service.ts
```diff
@@ -331,6 +101,78 @@ export class WebhookService extends BaseService {
     })
   }
 
+  private async handleInvoiceCreated(
```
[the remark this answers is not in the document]
SandipBajracharya: Yes, that's a good point. Moving the functions to their corresponding services makes it cleaner. That being said, there could be lots of changes, and it might need some more time for refactoring. I guess we can put this for later? @arpandhakal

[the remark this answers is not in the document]
arpandhakal: Sounds good @SandipBajracharya

PR #127, opened 2025-08-13
OUT-2128: allow access when companyId has "default" value

SandipBajracharya:
## Changes

- [X] no access if parsed token has clientId or companyId with value other than 'default'

## Testing Criteria

The 401 unauthorized issue can be seen when token has companyId with value "default". Like in this case:

image

Implemented condition to by-pass this case.

### DIFF
--- a/src/app/api/core/utils/authenticate.ts
+++ b/src/app/api/core/utils/authenticate.ts
@@ -16,7 +16,10 @@ export const _authenticateWithToken = async (token: string): Promise<User> => {
   }
 
   // Access to IU only. The reason for not checking the IUID directly is for the webhook events to pass
-  if (payload.data.clientId || payload.data.companyId) {
+  if (
+    payload.data.clientId ||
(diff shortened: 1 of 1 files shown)


src/app/api/core/utils/authenticate.ts
```diff
@@ -16,7 +16,12 @@ export const _authenticateWithToken = async (token: string): Promise<User> => {
   }
 
   // Access to IU only. The reason for not checking the IUID directly is for the webhook events to pass
-  if (payload.data.clientId || payload.data.companyId) {
+  if (
+    payload.data.clientId ||
+    (payload.data.companyId &&
+      payload.data.internalUserId &&
+      payload.data.companyId != 'default') // allow case when the parsed token have companyId = "default"
+  ) {
```
arpandhakal, line 24:
```suggestion
  if (
   !payload.data.internalUserId
  ) {
```

I think this works, doesn't it?

PR #129, opened 2025-08-14
OUT-2182: Issue when invoice fee is paid by client

SandipBajracharya:
## Changes

- [X] condition check for paid invoice. Only calculate client fees paid if the invoice is paid


rrojan (approved):
LGTM 🏆

PR #141, opened 2025-08-19
OUT-2211: Get type = "Service" items from QuickBooks for mapping

SandipBajracharya:
## Changes

- [X] only fetch "Service" type items from QB. Other type like "Category" cannot be used in invoice line items and throws error.


rrojan (approved):
LGTM 🏆

PR #155, opened 2025-09-05
OUT_2309 | Hotfix: add assembly domains to CSP

rrojan:
## Changes

- [x] Add assembly domains to be allowed in CSP


SandipBajracharya:
LGTM 🙌

PR #215, opened 2026-04-09
OUT-3542: add Sentry context tags, breadcrumbs, and error categorization

SandipBajracharya:
## Summary
- Add `portalId`, `workspaceId`, `entityType`, `eventType`, `errorCategory` tags to all Sentry error reports for dashboard filtering
- Add sync breadcrumbs at key decision points (invoice creation, customer search/creation, product mapping, payment processing) to trace execution paths leading to errors
- Expand `FailedRecordCategoryType` with `RATE_LIMIT`, `VALIDATION`, `QB_API_ERROR`, `MAPPING_NOT_FOUND` for finer error categorization
- Enhance `withErrorHandler` to capture errors with category and source tags via `Sentry.withScope`
- Add error source tracking (`intuit`/`copilot`) to `getMessageAndCodeFromError` for accurate QB API error detection
- Isolate Sentry scope per portal in Trigger.dev cron flow (`CronService._scheduleSinglePortal`) to prevent breadcrumb/tag bleed across portals
- Enrich threshold `captureMessage` in `checkAndUpdateAttempt` with full context tags (`portalId`, `entityType`, `eventType`, `errorCategory`) and extras (`invoiceNumber`, `errorMessage`, `attempt`)
- Fix ANSI color codes (`[32m`/`[39m`) leaking into Vercel/New Relic logs by disabling colors in non-dev environments

## DB Migration
Adds 4 new values to `failed_record_category_types` PostgreSQL enum:
```sql
ALTER TYPE "public"."failed_record_category_types" ADD VALUE IF NOT EXISTS 'rate_limit';
ALTER TYPE "public"."failed_record_category_types" ADD VALUE IF NOT EXISTS 'validation';
ALTER TYPE "public"."failed_record_category_types" ADD VALUE IF NOT EXISTS 'qb_api_error';
ALTER TYPE "public"."failed_record_category_types" ADD VALUE IF NOT EXISTS 'mapping_not_found';
```
Additive only — existing records and the `others` default are unaffected.

## Test plan
- [x] Verify `yarn build` passes
- [x] Run migration on staging DB, verify `SELECT enum_range(NULL::failed_record_category_types)` returns all 7 values
- [ ] Trigger a test webhook and verify Sentry event includes `portalId` and `entityType` tags
- [ ] Trigger a failed sync retry up to threshold and verify `captureMessage` includes enriched tags/extras
- [ ] Verify Vercel logs no longer contain ANSI escape codes
- [ ] Verify breadcrumbs appear in Sentry event trail for invoice/customer/product/payment flows

### DIFF
--- a/src/app/api/quickbooks/cron/cron.service.ts
+++ b/src/app/api/quickbooks/cron/cron.service.ts
@@ -10,6 +10,7 @@ import { getAllActivePortalConnections } from '@/db/service/token.service'
 import { CopilotAPI } from '@/utils/copilotAPI'
 import { encodePayload } from '@/utils/crypto'
 import CustomLogger from '@/utils/logger'
+import * as Sentry from '@sentry/nextjs'
 import dayjs from 'dayjs'
 import { and, eq, lt } from 'drizzle-orm'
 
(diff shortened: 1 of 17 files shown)


src/app/api/quickbooks/cron/cron.service.ts
```diff
@@ -18,29 +19,34 @@ export default class CronService {
...
-    const syncService = new SyncService(user)
+      // check if token is valid or not
+      const copilot = new CopilotAPI(token)
+      const tokenPayload = await copilot.getTokenPayload()
+      CustomLogger.info({
+        obj: { copilotApiCronToken: token, tokenPayload },
+        message:
+          'CronService#_scheduleSinglePortal | Copilot API token and payload',
+      })
+      if (!tokenPayload) throw new APIError(500, 'Encoded token is not valid') // this should trigger p-retry and re-run the function
```
priosshrsth, line 39: What we are doing here seems so wrong. I think we either set process.env.ASSEMBLY_ENV=local before line 32 and unset it after line 32. Or just use endpoints without sdks for cron. But looks like our whole flow is setup to require token. So I guess we are stuck with it.

src/utils/logger.ts
```diff
@@ -1,15 +1,17 @@
 import util from 'util'
 
+const isDev = process.env.NODE_ENV === 'development'
+
 const CustomLogger = {
   info({ message, obj }: { message: string; obj?: any }) {
     const consoleBody = [message]
-    if (obj) consoleBody.push(util.inspect(obj, { depth: null, colors: true }))
+    if (obj) consoleBody.push(util.inspect(obj, { depth: null, colors: isDev }))
```
priosshrsth, line 8: Any reason for this?

priosshrsth (approved):
@SandipBajracharya couple of comments. But overall good job.

src/app/api/quickbooks/cron/cron.service.ts
SandipBajracharya: Yes, we were using SDK from very beginning. To not use sdk, we will have to change most of the flows.

src/utils/logger.ts
SandipBajracharya: This is done to remove the ANSI escape codes from the log like shown in the following image. Those codes (`[32m`, `[39m`) are specially for colors which only makes sense in dev environment.

image

PR #216, opened 2026-04-09
OUT-3539: extract findOrMapInvoiceFromQBO helper for invoice sync resilience

SandipBajracharya:
## Summary
- Adds `findOrMapInvoiceFromQBO()` helper that queries QBO by invoice number, resolves/creates the customer mapping, and inserts a `qb_invoice_sync` row when found
- Integrates the helper into all 4 webhook handlers (`created`, `paid`, `voided`, `deleted`) as a fallback when no local mapping exists
- Refactors `checkIfInvoiceExistsInQBO()` to delegate to the new helper, so existing callers (`syncMissedInvoices`, `backfillTimedOutInvoices`) automatically gain mapping creation

## Problem
When an invoice is manually created in QBO, our app tries to create the same invoice and fails. Subsequent `invoice.paid`, `invoice.void`, and `invoice.delete` events also silently skip because no `qb_invoice_sync` mapping row exists in our database.

## Test plan
- [x] Invoice exists in QBO but not in local DB → `created` webhook maps it and skips creation
- [x] `paid` / `voided` / `deleted` webhooks with unmapped invoice → maps from QBO and proceeds
- [x] Invoice already mapped locally → handlers continue as before (no extra QBO API call)
- [x] Invoice not in QBO at all → handlers silently return as before
- [x] Calling `findOrMapInvoiceFromQBO` twice for the same invoice does not create duplicates

### DIFF
--- a/src/app/api/quickbooks/invoice/invoice.service.ts
+++ b/src/app/api/quickbooks/invoice/invoice.service.ts
@@ -557,6 +557,7 @@ export class InvoiceService extends BaseService {
       console.info(
         'InvoiceService#handleWebhookEvent#exists | Invoice already exists in the db',
       )
+
       return
     }
 
@@ -863,10 +864,10 @@ export class InvoiceService extends BaseService {
         intuitApi,
       })
       if (!mappedInvoice) {
-        console.info(
-          'InvoiceService#webhookInvoicePaid | Invoice not found in QBO either. Skipping.',
+        throw new APIError(
+          httpStatus.NOT_FOUND,
+          `Invoice not found in sync table or QBO for paid event. Invoice number: ${payload.data.number}. Likely preceded by a failed CREATE sync.`,
         )
-        return
       }
       invoiceSync = mappedInvoice
     }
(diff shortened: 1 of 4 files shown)


src/app/api/quickbooks/invoice/invoice.service.ts
```diff
@@ -1164,41 +1280,125 @@ export class InvoiceService extends BaseService {
...
         qbInvoiceId: qbInvoice.Id,
-        invoiceNumber: invoice.number,
+        invoiceNumber,
       },
       EventType.CREATED,
       {
-        amount: (invoice.lineItems[0].amount * 100).toFixed(2),
-        taxAmount: invoice.taxAmount ? invoice.taxAmount.toFixed(2) : '0',
+        amount: total ? total.toFixed(2) : undefined,
+        taxAmount: taxAmount ? taxAmount.toFixed(2) : '0',
```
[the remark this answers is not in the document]
SandipBajracharya: `total` is already in cents.

src/app/api/quickbooks/sync/sync.service.ts
```diff
@@ -177,7 +177,7 @@ export class SyncService extends BaseService {
       const invoice = {
         id: record.copilotId,
         number: invNumber,
-        total: record.amount ? parseFloat(record.amount) / 100 : 0, // assuming amount is in cents
+        total: record.amount ? parseFloat(record.amount) : 0,
```
priosshrsth, line 180: Won't this cause regression?

priosshrsth (changes requested):

src/app/api/quickbooks/sync/sync.service.ts
SandipBajracharya: This shouldn’t cause any regression. The change is only in the invoice void/delete function, and we’re not actually using the total amount for syncing anyway—it’s just there for logging.

priosshrsth (approved):
lgtm

PR #217, opened 2026-04-09
OUT-3546: set up Vitest with initial unit tests for critical utils

SandipBajracharya:
## Summary

Add Vitest infrastructure with test scripts (test, test:watch, test:coverage), v8 coverage config, and CI workflow running tests on pull requests. Include unit tests for string utils (escapeForQBQuery, replaceSpecialCharsForQB) and withRetry (429 retry, RetryableError, exhaustion) with fake timers.


priosshrsth (approved):
@SandipBajracharya PR looks good to me. I would suggest to keep all tests in a single directory. Because as we increase the tests, it'd only pollute our source codes. You can do that in separate PR if you'd like. And in ci, you could run linting as well to run in parallel with test.

SandipBajracharya:
> @SandipBajracharya PR looks good to me. I would suggest to keep all tests in a single directory. Because as we increase the tests, it'd only pollute our source codes. You can do that in separate PR if you'd like. And in ci, you could run linting as well to run in parallel with test.

We have tests in single parent directory named `tests`. I created sub directories in `tests` folder to categorize the tests itself. We already have separate CI that runs linting on every push.

PR #218, opened 2026-04-10
fix(OUT-3575): handle Intuit OAuth's new error format for auth errors

SandipBajracharya:
## Summary
- Intuit silently changed their OAuth error response from Axios-style `{ response: { status, data } }` to `{ error, error_description, intuit_tid }`. This caused refresh token expiry errors to go undetected, preventing notification emails to IUs.
- Added `IntuitOAuthError extends Error` class with a `fromRaw()` factory method to wrap Intuit's plain-object errors into proper Error instances. This is necessary because `pRetry` discards non-Error throws.
- Added `isIntuitOAuthError` type guard (now `instanceof` based) and updated all error handling paths.
- Extracted `OAuthErrorCodes.INVALID_GRANT` constant to replace magic string comparisons.

## Changes
| File | Change |
|---|---|
| `src/app/api/core/exceptions/custom.ts` | `IntuitOAuthError` class with `fromRaw()` factory + `isIntuitOAuthError` instanceof guard |
| `src/app/api/core/utils/withErrorHandler.ts` | Handle `IntuitOAuthError` before generic `Error` branch + Sentry capture |
| `src/app/api/quickbooks/auth/auth.service.ts` | Switch from `isAxiosError` to `isIntuitOAuthError`, use constant for `invalid_grant` check |
| `src/constant/intuitErrorCode.ts` | Added `OAuthErrorCodes` with `as const` |
| `src/utils/error.ts` | Handle `IntuitOAuthError` before generic `Error` in `getMessageAndCodeFromError` |
| `src/utils/intuit.ts` | Wrap plain Intuit OAuth errors via `IntuitOAuthError.fromRaw()` in all SDK calls (`_refreshAccessToken`, `_authorizeUri`, `_createToken`) |

## Key design decisions
- **Why `IntuitOAuthError extends Error`?** — `pRetry` discards non-Error throws with a generic `TypeError`, losing all Intuit error fields. Wrapping in a proper Error subclass preserves them through the retry pipeline.
- **Why `fromRaw()` factory?** — Keeps plain-object shape detection in one place. Callers just do `throw IntuitOAuthError.fromRaw(error) ?? error`.
- **Why reorder `isIntuitOAuthError` before `instanceof Error`?** — Since `IntuitOAuthError extends Error`, the generic `instanceof Error` branch would catch it first, making our specific handler dead code.
- **No retry for OAuth errors** — `shouldRetry` in `withRetry.ts` only retries `status === 429`. `IntuitOAuthError` has no status, so it fails fast. Network/rate-limit errors still retry normally.

## Test plan
- [ ] Simulate expired refresh token — verify `turnOffSync` is called and IU notification email is sent
- [ ] Simulate non-`invalid_grant` OAuth error (e.g. `invalid_client`) — verify error is caught and reported to Sentry
- [ ] Verify existing Axios-style errors from QuickBooks Data Services API are still handled correctly
- [ ] Verify network errors during token refresh are still retried (up to 3 times)

### DIFF
--- a/src/app/api/core/exceptions/custom.ts
+++ b/src/app/api/core/exceptions/custom.ts
@@ -14,3 +14,22 @@ export function isAxiosError(
     'status' in (error as any).response
   )
 }
+
+/**
+ * Intuit-Oauth has its own error format response.
+ * Format: { error: string, error_description: string, intuit_tid: string }
+ */
+export function isIntuitOAuthError(
+  error: unknown,
+): error is { error: string; error_description: string; intuit_tid: string } {
+  return (
+    typeof error === 'object' &&
+    error !== null &&
+    'error' in error &&
+    typeof (error as any).error === 'string' &&
+    'error_description' in error &&
+    typeof (error as any).error_description === 'string' &&
+    'intuit_tid' in error &&
+    typeof (error as any).intuit_tid === 'string'
+  )
+}
--- a/src/app/api/core/exceptions/custom.ts
+++ b/src/app/api/core/exceptions/custom.ts
@@ -18,18 +18,47 @@ export function isAxiosError(
 /**
  * Intuit-Oauth has its own error format response.
  * Format: { error: string, error_description: string, intuit_tid: string }
+ *
+ * Wrapping in an Error subclass so pRetry doesn't discard the original fields.
  */
-export function isIntuitOAuthError(
-  error: unknown,
-): error is { error: string; error_description: string; intuit_tid: string } {
-  return (
-    typeof error === 'object' &&
-    error !== null &&
-    'error' in error &&
-    typeof (error as any).error === 'string' &&
-    'error_description' in error &&
-    typeof (error as any).error_description === 'string' &&
-    'intuit_tid' in error &&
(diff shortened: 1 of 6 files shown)


src/app/api/core/exceptions/custom.ts
```diff
@@ -14,3 +14,22 @@ export function isAxiosError(
...
+  return (
+    typeof error === 'object' &&
+    error !== null &&
+    'error' in error &&
+    typeof (error as any).error === 'string' &&
+    'error_description' in error &&
+    typeof (error as any).error_description === 'string' &&
+    'intuit_tid' in error &&
+    typeof (error as any).intuit_tid === 'string'
+  )
```
priosshrsth: This is bad. Can we simplify this.

```suggestion
  return (
    if (!error || typeof error !== 'object') {
    return false;
    }
    if((!('intuid_tid' in error)) {
    return false
    }
    
    ....
    ....
    ...
```

priosshrsth (changes requested):
@SandipBajracharya Could you simplify the function to check intuit error?

priosshrsth (approved):
lgtm

PR #219, opened 2026-04-10
OUT-3578: add afterIfAvailable utility for non-request contexts

SandipBajracharya:
## Summary
- Next.js `after()` throws when called outside a request scope (Trigger.dev background jobs, CLI scripts).
- Added `afterIfAvailable()` utility that tries `after()` and falls back to direct async execution when no request context is available.
- Applied to `getQBPortalConnection()`'s notification path in `auth.service.ts`, which runs in both Vercel serverless and Trigger.dev contexts.

## Changes
| File | Change |
|---|---|
| `src/app/api/core/utils/afterIfAvailable.ts` | New utility — tries `after()`, catches scope errors, falls back to direct execution with `.catch()` error logging |
| `src/app/api/quickbooks/auth/auth.service.ts` | Use `afterIfAvailable` for IU notification in `getQBPortalConnection`; `handleTokenExchange` keeps `after()` (request-only) |

## Test plan
- Trigger a refresh token expiry via Trigger.dev `resyncFailedRecords` task — verify notification fires without `after()` scope error
- Trigger a refresh token expiry via Vercel serverless route — verify `after()` still defers the callback as before
- Simulate notification failure in fallback path — verify error is logged via `[afterIfAvailable] callback failed:`

### DIFF
--- a/src/app/api/core/utils/afterIfAvailable.ts
+++ b/src/app/api/core/utils/afterIfAvailable.ts
@@ -1,0 +1,21 @@
+import { after } from 'next/server'
+
+/**
+ * Runs the callback via Next.js `after()` if inside a request scope,
+ * otherwise executes it directly. This allows shared service code to
+ * work in both Vercel serverless (request context) and Trigger.dev /
+ * CLI (no request context) environments.
+ *
+ * `after()` is a pure registration call — the only errors it can throw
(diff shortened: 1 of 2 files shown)


priosshrsth (approved):
lgtm. Good to know. I have added after in all of repos as well.

PR #223, opened 2026-04-16
OUT-3591: truncate long service names for QBO and warn users

SandipBajracharya:
## Summary
- Removes `varchar(100)` constraint from `name` and `copilotName` columns in `qb_product_sync` and `remark`/`productName` in `qb_sync_logs`
- Adds `truncateForQB()` utility that truncates names to QBO's 100-char limit with ellipsis ("..."), preserving duplicate-name suffixes like ` (N)`
- Applies truncation at all QBO API call sites (item create, update, invoice sync, missed products sync)
- Shows a warning `Callout` and per-row warning icons on the product mapping UI when service names exceed 95 chars (buffer for suffixes)
- Replaces bare `console.error` with `CustomLogger.error` in webhook handlers

## Test plan
- [x] Create a service in Assembly with a name > 100 characters
- [x] Verify warning callout and icon appear on the product mapping page
- [x] Trigger a product sync and verify the item is created in QBO with a truncated name ending in "..."
- [x] Verify duplicate product names get the ` (N)` suffix preserved after truncation
- [x] Verify services with names <= 95 chars show no warning
- [x] Verify the DB migration runs without issues

## Testing Criteria

https://www.loom.com/share/b4d3101cf7ee4be79bc958ff53c761f4


src/db/schema/qbSyncLogs.ts
```diff
@@ -33,12 +33,12 @@ export const QBSyncLog = table('qb_sync_logs', {
   quickbooksId: t.varchar('quickbooks_id', { length: 100 }),
   invoiceNumber: t.varchar('invoice_number', { length: 100 }),
   amount: t.decimal('amount'),
-  remark: t.varchar('remark', { length: 255 }),
+  remark: t.varchar('remark'),
```
priosshrsth, line 36: @SandipBajracharya What was the reason of removing it? Also I think varchar has default limit of 256 characters.

priosshrsth (changes requested):

src/db/schema/qbSyncLogs.ts
SandipBajracharya: varchar without limit is equivalent to text. It can store long characters.

priosshrsth (approved):

PR #227, opened 2026-04-17
OUT-3604: QBO bank deposits for automatic bank reconciliation

SandipBajracharya:
## Summary

- When payments are received, optionally create QBO Bank Deposits that match the net amount deposited to the customer's bank (after Stripe fees), enabling automatic bank transaction matching
- Payments are routed through Undeposited Funds, then a Bank Deposit moves the net amount to the bank account with the fee recorded as a negative line
- New opt-in `bankDepositFeeFlag` setting (requires `absorbedFeeFlag` to also be enabled)
- New `GET /api/quickbooks/setting/bank-account` endpoint to fetch QBO bank accounts
- Settings POST now handles `bankAccountRef` write to `qb_portal_connections` atomically with flag updates
- Fix: missing `undepositedFundsAccountRef`/`bankAccountRef` in token refresh path
- Fix: Undeposited Funds lookup uses `AccountSubType` query (not getOrCreate)
- Fix: deposit payload includes required `TxnLineId`
- Fix: null guard on `paidSyncLog.amount` before deposit creation
- Sentry breadcrumbs and structured logging across the bank deposit flow

## Test plan

- [ ] Enable `absorbedFeeFlag` + `bankDepositFeeFlag` in settings
- [ ] Select a bank account from the dropdown
- [ ] Trigger an invoice payment through Stripe
- [ ] Verify `invoice.paid` webhook routes QBO Payment to Undeposited Funds
- [ ] Verify `payment.succeeded` webhook creates a Bank Deposit (gross - fee = net)
- [ ] Verify QBO Bank Deposit matches the real bank feed amount
- [ ] Verify disabling flags reverts to legacy expense flow
- [ ] Verify existing payments (without flag) are unaffected


src/app/api/quickbooks/webhook/webhook.service.ts
```diff
@@ -396,7 +399,7 @@ export class WebhookService extends BaseService {
     payload: unknown,
     qbTokenInfo: IntuitAPITokensType,
   ) {
-    await sleep(7000) // Payment succeed event can sometimes trigger before invoice created.
+    await sleep(20000) // Payment succeed event can sometimes trigger before invoice created.
```
arpandhakal, line 402: 7 seconds to 20 seconds is a huge bump. does this effect the performance of the application?

src/type/common.ts
```diff
@@ -288,6 +290,13 @@ export const SettingRequestSchema = z
           message: 'absorbedFeeFlag is required when type is invoice',
         })
       }
+      if (typeof val.bankDepositFeeFlag !== 'boolean') {
+        ctx.addIssue({
+          path: ['bankDepositFeeFlag'],
+          code: z.ZodIssueCode.custom,
+          message: 'bankDepositFeeFlag is required when type is invoice',
+        })
+      }
```
arpandhakal, line 299: dont we need extra validation here? bankDepositFeeFlag is dependent on absorbedFeeFlag and bankAccountRef is dependent on bankDepositFeeFlag. We can add a validation  here so that bankDepositFeeFlag requires absorbedFeeFlag, and bankAccountRef requires bankDepositFeeFlag.

arpandhakal (approved):

src/app/api/quickbooks/webhook/webhook.service.ts
SandipBajracharya: New deposit flow requires invoice to be paid otherwise the flow fails. Bumped sleep to 20 sec to be very sure that `invoice.paid` event is complete. Since those syncs run at background, I think we have to compromise speed with accuracy.

src/type/common.ts
SandipBajracharya: Correct but for now we are only releasing this feature to specific portals so adding validation here might make things more rigid. Instead we are handling this case in controller for now after checking if the portal accessing the resource is eligible for the feature

PR #229, opened 2026-04-20
OUT-3617: AB testing gate for bank deposit feature

SandipBajracharya:
## Summary

- Add `AB_FEATURE_TESTING_PORTALS` env var for incremental rollout of bank deposit feature
- Portals in the list get the bank deposit UI and webhook flow; others use the legacy expense flow
- Empty/unset env var = feature available to all (production rollout state)

## Changes

- `src/config/index.ts` — parse `AB_FEATURE_TESTING_PORTALS` comma-separated list
- `src/utils/abTesting.ts` — `isPortalInBankDepositABTest(portalId)` utility
- `src/app/api/quickbooks/invoice/invoice.service.ts` — AB check on `useBankDepositFlow`
- `src/app/api/quickbooks/webhook/webhook.service.ts` — AB check on `useBankDepositFlow`
- `src/app/api/quickbooks/setting/setting.controller.ts` — return `bankDepositEnabled` in GET
- `src/hook/useSettings.ts` — conditional bank account fetch based on `bankDepositEnabled`
- `src/components/` — hide deposit checkbox + dropdown when portal not in AB test
- Remove `EventType.DEPOSITED` — use `SUCCEEDED` for both flows (flag decides behavior)
- Fix cascade bug: only reset `bankDepositFeeFlag` when feature is visible

## Test plan

- [ ] Set `AB_FEATURE_TESTING_PORTALS=<portal-id>` in `.env`
- [ ] Portal in list: deposit checkbox, dropdown, and bank deposit flow work
- [ ] Portal not in list: checkbox hidden, legacy expense flow used
- [ ] Toggle absorbed fees on/off for non-AB portal — buttons disappear correctly
- [ ] Remove env var entirely — feature available to all portals


arpandhakal (commented):
The migrations is being rewritten and not updated. I see the old migration file deleted and new one is introduced. Are these already pushed into preview, production databases? If so, I think that will cause a problem. Flagging this early.

SandipBajracharya:
> The migrations is being rewritten and not updated. I see the old migration file deleted and new one is introduced. Are these already pushed into preview, production databases? If so, I think that will cause a problem. Flagging this early.

No we have not pushed those migration to production yet. However, we have ran this in supabase dev. It makes sense to update the migration file in first PR i.e on [OUT-3064](https://github.com/assemblycom/quickbooks-sync/pull/227). Will update this and let you know

PR #231, opened 2026-04-21
OUT-3621: proactive token refresh with race-safe invalid_grant handling

SandipBajracharya:
## Summary
- Replaces two duplicated `isExpired` blocks with a single `getValidQbTokens(portalId)` helper that proactively refreshes when the stored token is within `REFRESH_BUFFER_SECONDS` (15 min) of expiry.
- On `invalid_grant`, `getRefreshedQbTokenInfo` distinguishes a cross-worker refresh race (DB `tokenSetTime` advanced) from genuine revocation by re-reading the row — clock-skew-free.
- Genuine revocation throws a typed `QBReconnectRequiredError` without mutating state. `AuthService.getQBPortalConnection` owns the paired disable-sync + notify-IU side effects so they only fire under a Copilot user context.

## Test plan
- [x] `yarn test` — 50 tests across 3 files, all passing (10 new in `tokenRefresh.test.ts`).
- [x] `yarn tsc --noEmit` — clean (pre-existing unrelated errors filtered).
- [x] Manual: connect a sandbox portal, force an expired token (or wait 1h), confirm the next webhook proactively refreshes without 401 noise.
- [x] Manual: revoke a sandbox portal in Intuit admin, fire a webhook, confirm `syncFlag` flips to false and the IU receives the reconnect notification once.
- [x] Manual: load the dashboard for a portal whose token is revoked — UI should not crash; `checkForNonUsCompany` returns false silently.

## Linear
[OUT-3621](https://linear.app/assemblycom/issue/OUT-3621/intuitapi-silent-401-proactive-token-refresh-with-race-safe-handling)


priosshrsth (approved):
lgtm

PR #233, opened 2026-04-21
OUT-3625: suffix DisplayName on QB Vendor/Employee collision

SandipBajracharya:
## Summary
- Fixes QB customer creation failing with error 6240 when the Copilot client's `DisplayName` collides with an existing QB Vendor or Employee (QBO enforces `DisplayName` uniqueness across all three entity types).
- Adds a pre-check via a new `IntuitAPI.getNameCollisionEntity(displayName)` that queries Vendor then Employee; if a collision is found, appends ` (Customer)` suffix to the payload `DisplayName` before calling `createCustomer`.
- Separate refactor commit removes 8 unreachable Fault-check branches that sat after `customQuery` calls (dead because `_customQuery` already throws on Fault and returns `res.QueryResponse`, which has no `Fault` field).

## Notes
- Inactive Vendors/Employees are intentionally excluded from the collision check — QBO auto-suffixes their `DisplayName` with ` (deleted)` when deactivated, freeing the original name.
- `_getNameCollisionEntity` short-circuits after finding a Vendor match (skips the Employee query).
- Mapping table's `displayName` column continues to store the Copilot-side original (not the suffixed QB name); the column's semantic is preserved and a comment now documents the intentional divergence.
- Critical-path tests were not added (no test infrastructure in this repo yet per OUT-3625 memory; separate initiative).

## Test plan
- [ ] In QB sandbox: create a Vendor named `Acme Corp`, then trigger a Copilot invoice webhook for a client whose display name resolves to `Acme Corp` → verify a Customer named `Acme Corp (Customer)` is created in QB without error 6240.
- [ ] In QB sandbox: deactivate a Vendor named `Beta Inc` (QB renames it to `Beta Inc (deleted)`), trigger webhook for a client `Beta Inc` → verify a Customer named `Beta Inc` is created (no suffix, since the inactive vendor no longer holds the name).
- [ ] Happy path regression: trigger webhook for a recipient whose display name does not collide with any Vendor/Employee → verify customer is created with the unsuffixed `DisplayName` and only 2 QB API calls were made (Vendor query + Employee query, or fewer if short-circuit).
- [ ] Verify existing `findOrCreateCustomer` behaviors unchanged (email-based short-circuit, `CompanyName` mismatch clearing).
- [ ] Refactor regression: trigger a query that would fault (e.g., invalid realm) → verify error surfaces with `IntuitAPIErrorMessage` prefix so `isIntuitError` detection in `src/utils/error.ts` still classifies it correctly.

### handling of duplicate name across Customer, Employee and Vendor

https://www.loom.com/share/14cdacf0e4ed4918ac5b9454ac918eb5

### Case when two companies have same name

https://www.loom.com/share/11c319b68ec54fee8b7c8be75ba57581

### DIFF
--- a/src/app/api/quickbooks/customer/customer.service.ts
+++ b/src/app/api/quickbooks/customer/customer.service.ts
@@ -16,7 +16,7 @@ import { InvoiceCreatedResponseType } from '@/type/dto/webhook.dto'
 import { CopilotAPI } from '@/utils/copilotAPI'
 import IntuitAPI from '@/utils/intuitAPI'
 import { addSyncBreadcrumb } from '@/utils/sentry'
-import { getNameAsCustomer, replaceSpecialCharsForQB } from '@/utils/string'
+import { replaceSpecialCharsForQB } from '@/utils/string'
 import { and, eq, isNull } from 'drizzle-orm'
 import httpStatus from 'http-status'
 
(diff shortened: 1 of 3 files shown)


src/app/api/quickbooks/customer/customer.service.ts
```diff
@@ -376,8 +376,24 @@ export class CustomerService extends BaseService {
...
+      const sanitizedDisplayName = replaceSpecialCharsForQB(displayName)
+
+      // QB enforces DisplayName uniqueness across Customer, Vendor, and Employee.
+      // If the name collides with a Vendor/Employee, suffix it to avoid error 6240.
+      const collision =
+        await intuitApiService.getNameCollisionEntity(sanitizedDisplayName)
+      const finalDisplayName = collision
+        ? getNameAsCustomer(sanitizedDisplayName)
+        : sanitizedDisplayName
+
```
priosshrsth: Could these be valid cases?
1. "Customer One" has this condition and we create "Customer One (customer)"
2. For some reason, this code again runs for the customer, but now same things happen but we cannot create "Customer One (customer)" because it was already created.

If a customer is mapped, the referece id is already saved in DB right? and this will not run twice for same customer. Otherwiese we might beed to get everythin matching "Customer one" and fin the customer by checking if it matches the sanitized on affixed version of the name.

priosshrsth (changes requested):

src/app/api/quickbooks/customer/customer.service.ts
SandipBajracharya: Once a customer is mapped, we wont be creating same customer again. However there can be case where two customer (invoice billed to company) has same name. Then this case might arise. I will push changes for that case.

priosshrsth (approved):
lgtm. I think istead of appending number to suffixed name, we could do it in displayName. But it looks good to me as it is as well.

PR #236, opened 2026-04-27
OUT-3644: add idempotency guards on qb_invoice_sync writes

SandipBajracharya:
## Summary

- Partial unique index on `qb_invoice_sync (portal_id, invoice_number) WHERE deleted_at IS NULL` + `onConflictDoNothing` in `createQBInvoice` — concurrent webhook deliveries can no longer insert duplicate mapping rows.
- Inline re-check guard in `findOrMapInvoiceFromQBO` after the QBO fetch; race-loss path skips redundant `logSync`.
- `getOneByCopilotIdAndEventType` and the conditions branch of `updateOrCreateQBSyncLog` now filter `isNull(deletedAt)` so soft-deleted log rows can't be revived in-place.

`qb_sync_logs` unique index was descoped — production has too many historical duplicates to add the constraint cleanly. Webhook-entry idempotency tracked in [OUT-3655](https://linear.app/assemblycom/issue/OUT-3655).

Linear: https://linear.app/assemblycom/issue/OUT-3644

## Test plan

- [ ] Apply migration in dev, confirm partial unique index exists on `qb_invoice_sync`
- [ ] Replay two `invoice.created` webhooks in parallel for the same invoice — verify only one `qb_invoice_sync` row, only one `qb_sync_logs` CREATED row
- [ ] Soft-delete a `qb_sync_logs` row, then trigger a sync that would normally update it — verify a new row is created instead of reviving the soft-deleted one
- [ ] Existing sync flows (create / paid / voided / deleted) regress-test against QBO sandbox

### DIFF
--- a/src/app/api/quickbooks/invoice/invoice.service.ts
+++ b/src/app/api/quickbooks/invoice/invoice.service.ts
@@ -785,7 +785,20 @@ export class InvoiceService extends BaseService {
       customerId: existingCustomerMapId, // foreign key to customer mapping
       status: invoiceResource.status,
     }
-    await this.createQBInvoice(invoicePayload)
+    const inserted = await this.createQBInvoice(invoicePayload, ['id'])
+
+    // If onConflictDoNothing skipped the insert, a concurrent delivery won
+    // the race. Skip logSync (the winner already wrote the CREATED log;
+    // overwriting it would point quickbooks_id at this losing webhook's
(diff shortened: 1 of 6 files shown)


priosshrsth (approved):
src/app/api/quickbooks/invoice/invoice.service.ts is becoming too large, don't you think?
Approved. But maybe we should also think about simplifying this in the future.

PR #240, opened 2026-05-05
OUT-3545: parallelize home render and batch Copilot price fetch

SandipBajracharya:
## Summary
- Parallelize portal connection check, workspace info fetch, and latest sync log lookup on the home server component (`src/app/(home)/Home.tsx`); replace internal API round-trip with direct `SyncLogService` call.
- Batch Copilot price fetching in `ProductService.getFlattenProductList` — paginate `/prices` once and group by `productId` instead of N+1 calls per product (drops `bottleneck`).
- Add a `loading.tsx` for the home segment and reuse it from `dashboard/Main`; drop the now-unused `checkSyncStatus` action (sync flag now read from the eagerly-loaded portal connection's setting).

## Test plan
- [x] Load the home route signed in to a portal that has an active QBO connection — verify spinner renders during streaming, then dashboard hydrates with the correct sync flag, enabled state, and last-sync timestamp.
- [x] Load the home route signed in to a portal with no QBO connection — verify connect CTA renders and `reconnect` flow still triggers when `?type=...` is passed.
- [x] Open the products page / any view that calls `getFlattenProductList` and confirm products with multiple prices still render every price row, sorted ascending by amount.
- [x] Confirm Copilot price pagination terminates: portal with >100 prices should still return all of them (paginate via `nextToken`).
- [x] `yarn lint:check && yarn test` clean.

### DIFF
--- a/src/app/api/quickbooks/product/product.service.ts
+++ b/src/app/api/quickbooks/product/product.service.ts
@@ -12,9 +12,8 @@ import {
   ProductChangedItemReferenceType,
   ProductMappingSchemaType,
 } from '@/db/schema/qbProductSync'
-import { ProductResponse, WhereClause } from '@/type/common'
+import { PriceResponse, WhereClause } from '@/type/common'
 import { ProductFlattenArrayResponseType } from '@/type/dto/api.dto'
-import { bottleneck } from '@/utils/bottleneck'
 import { QBItemFullUpdatePayloadType } from '@/type/dto/intuitAPI.dto'
 import {
   PriceCreatedResponseType,
@@ -41,6 +40,7 @@ import {
 } from '@/utils/string'
 import { AccountTypeObj } from '@/constant/qbConnection'
 import { TokenService } from '@/app/api/quickbooks/token/token.service'
+import { MAX_PRODUCT_LIST_LIMIT } from '@/app/api/core/constants/limit'
 
 export type ProductSyncTokenResponse = {
   id: string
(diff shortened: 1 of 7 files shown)


SandipBajracharya:
@greptileai review

src/app/api/quickbooks/product/product.service.ts
```diff
@@ -319,47 +319,73 @@ export class ProductService extends BaseService {
...
-        description: convert(product.description),
-        priceId: price.id,
-        amount: price.amount,
-        type: price.type,
-        interval: price.interval,
-        intervalCount: price.intervalCount,
-        currency: price.currency,
-      }))
-      .sort((a, b) => a.amount - b.amount) // sort by amount in asc order
+  async getFlattenProductList(): Promise<ProductFlattenArrayResponseType> {
```
priosshrsth:
```suggestion
  async getProductsWithPrices(): Promise<ProductFlattenArrayResponseType> {
```

or just getFlatProductList sounds more intuitive that getFlattenProdctList. wdyt?

priosshrsth (approved):
One minor optional suggestion. But the PR  looks good to me and good one. This should help a lot. You are free to merge it as it is if you want.

src/app/api/quickbooks/product/product.service.ts
SandipBajracharya: Suggested one makes more sense. Thanks.

PR #241, opened 2026-05-06
OUT-3686: paginate-and-filter QBO customer email lookup

SandipBajracharya:
## Summary

- QBO's `/query` parser silently mishandles RFC-legal special characters in `PrimaryEmailAddr` filters (confirmed for `+`, both `=` and `LIKE` literal forms fail), returning 0 rows even when a matching customer exists. `findOrCreateCustomer` therefore missed existing QBO customers with plus-aliased emails and created duplicates.
- Replaces the WHERE-clause email filter with a paginated walk + JS-side match so the email never appears in the query — works for any RFC-legal address. Adds a `sanitizedCompanyName` parameter to disambiguate customers sharing an email across companies (one Copilot client can be enrolled in multiple companies); this absorbs the post-filter that previously did the same check in `customer.service.ts`.
- Outer `wrapWithRetry` intentionally dropped on `getCustomerByEmail` so a mid-walk 429 retries only the failing page, not the entire realm walk from page 1. Inner `customQuery` retries are preserved.

## Test plan

- [x] 21 new unit tests in `test/unit/utils/intuitAPI.test.ts` covering pagination termination, case/whitespace normalisation, malformed `PrimaryEmailAddr` shapes, off-by-one full-page-then-empty edge case, the `+`-alias regression, and company-aware matching with multi-page walks
- [x] `yarn tsc --noEmit` clean
- [x] `yarn lint:check` clean (0 errors)
- [x] Manual sandbox verification: create QBO customer with `user+tag@example.com`, trigger an invoice webhook, confirm `findOrCreateCustomer` finds the existing customer instead of creating a duplicate
- [x] Manual sandbox verification: same email across two QBO customers with different `CompanyName` — confirm walker returns the matching-company customer, not the first email-match

### DIFF
--- a/src/utils/intuitAPI.ts
+++ b/src/utils/intuitAPI.ts
@@ -296,20 +296,49 @@ export default class IntuitAPI {
     return CustomerQueryResponseSchema.parse(qbCustomers.Customer[0])
   }
 
+  // QBO's query parser silently mishandles certain special characters (confirmed
+  // for '+', and '=' / 'LIKE' literal both fail) when filtering on
+  // PrimaryEmailAddr, returning 0 results even when a matching customer exists.
+  // To stay correct for any RFC-legal email, we never put the email in the WHERE
+  // clause: page through customers and match client-side instead.
+  //
+  // sanitizedCompanyName disambiguates customers sharing the same email across
+  // companies (one Copilot client can be enrolled in multiple companies). The
+  // CompanyName comparison uses the same `(value || undefined)` normalisation
+  // as the post-filter in customer.service.ts so the two layers cannot disagree.
   async _getCustomerByEmail(
     email: string,
+    sanitizedCompanyName: string | undefined,
   ): Promise<CustomerQueryResponseType | undefined> {
+    const needle = email.trim().toLowerCase()
+    if (!needle) return
+
     CustomLogger.info({
-      obj: { email },
+      obj: { email, sanitizedCompanyName },
       message: `IntuitAPI#getCustomerByEmail | Customer query start for realmId: ${this.tokens.intuitRealmId}. Email: ${email}`,
     })
-    const customerQuery = `SELECT Id, SyncToken, Active, CompanyName, PrimaryEmailAddr FROM Customer WHERE PrimaryEmailAddr = '${escapeForQBQuery(email)}' AND Active in (true, false)`
-    const qbCustomers = await this.customQuery(customerQuery)
 
-    if (!qbCustomers) return
+    const pageSize = 1000
+    let startPosition = 1
 
-    if (!qbCustomers.Customer) return
-    return CustomerQueryResponseSchema.parse(qbCustomers.Customer[0])
+    while (true) {
+      const customerQuery = `SELECT Id, SyncToken, Active, CompanyName, PrimaryEmailAddr FROM Customer WHERE Active IN (true, false) STARTPOSITION ${startPosition} MAXRESULTS ${pageSize}`
+      const qbCustomers = await this.customQuery(customerQuery)
+      const customers = qbCustomers?.Customer ?? []
+      if (customers.length === 0) return
+
+      const match = customers.find((c: CustomerQueryResponseType) => {
+        const addr = c.PrimaryEmailAddr?.Address
+        if (typeof addr !== 'string') return false
+        if (addr.trim().toLowerCase() !== needle) return false
+        if ((c.CompanyName || undefined) !== sanitizedCompanyName) return false
+        return true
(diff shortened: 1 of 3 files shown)


src/utils/intuitAPI.ts
```diff
@@ -296,20 +296,49 @@ export default class IntuitAPI {
...
+        if (typeof addr !== 'string') return false
+        if (addr.trim().toLowerCase() !== needle) return false
+        if ((c.CompanyName || undefined) !== sanitizedCompanyName) return false
+        return true
+      })
+      if (match) return CustomerQueryResponseSchema.parse(match)
+
+      if (customers.length < pageSize) return
+      startPosition += pageSize
+    }
```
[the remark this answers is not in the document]
SandipBajracharya: Updated with `ORDERBY MetaData.CreatedAt Desc` instead by id. This way it is more performative

src/utils/intuitAPI.ts
```diff
@@ -296,20 +296,49 @@ export default class IntuitAPI {
...
   }
```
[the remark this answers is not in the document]
SandipBajracharya: This can be deferred. 100k customers in a portal is rare case. We can work at this as scale improvement.

src/utils/intuitAPI.ts
```diff
@@ -296,20 +296,61 @@ export default class IntuitAPI {
...
+  // so concurrent updates do not move rows and any customer created during the
+  // walk lands at the end of the cursor where we'll still encounter it.
+  //
+  // Tradeoff vs. CreateTime DESC: newly-created customers land on the LAST
+  // page rather than page 1, so drift recovery for a fresh customer in a 10k
+  // realm walks all pages (~5s) instead of hitting on page 1 (~500ms). The
+  // perf cost is bounded and acceptable; full stability is the priority.
   async _getCustomerByEmail(
     email: string,
+    sanitizedCompanyName: string | undefined,
```
priosshrsth:
```suggestion
    sanitizedCompanyName?: string,
```

src/utils/intuitAPI.ts
```diff
@@ -296,20 +296,61 @@ export default class IntuitAPI {
...
+      const customerQuery = `SELECT Id, SyncToken, Active, CompanyName, PrimaryEmailAddr FROM Customer WHERE Active IN (true, false) ORDERBY Id ASC STARTPOSITION ${startPosition} MAXRESULTS ${pageSize}`
+      const qbCustomers = await this.customQuery(customerQuery)
+      const customers = qbCustomers?.Customer ?? []
+      if (customers.length === 0) return
+
+      const match = customers.find((c: CustomerQueryResponseType) => {
+        const addr = c.PrimaryEmailAddr?.Address
+        if (typeof addr !== 'string') return false
+        if (addr.trim().toLowerCase() !== needle) return false
+        if ((c.CompanyName || undefined) !== sanitizedCompanyName) return false
```
priosshrsth, line 329:
```suggestion
        if (c.CompanyName !== sanitizedCompanyName) return false
```

shouldn't this be enough?

src/utils/intuitAPI.ts
```diff
@@ -296,20 +296,61 @@ export default class IntuitAPI {
...
+        return true
```
priosshrsth, line 330:
```suggestion
        return c.CompanyName === sanitizedCompanyName
```

priosshrsth (changes requested):
@SandipBajracharya Minor feedback. Otherwise everything looks good to me.

src/utils/intuitAPI.ts
SandipBajracharya: CompanyName can be empty string when returned from QBO. Since the value for sanitizedCompanyName is either string or optional, comparing CompanyName with empty string is never equivalent to undefined sanitizedCompanyName. That is the reason for such implementation.

priosshrsth (approved):
lgtm

src/utils/intuitAPI.ts
priosshrsth: This is confusing tbh. So if both are undefined === undefined or "" === "" vayo vane ni true nai janu parne haina?

Your check right now can have following scenarios:
undefined === string | undefined
`${string}` === string | undedinfed,

So only thing you want to avoid is `"" === string | undefined`?

SandipBajracharya: What I want to be true is "" === undefined. Left hand side is never undefined without `|| undefined`. QBO sends empty string when there is no value. So empty string becomes undefined with help of the logical OR operation.

PR #242, opened 2026-05-06
OUT-3544: Add retry mechanisms for transient network errors

SandipBajracharya:
## Summary

- Expand `withRetry` classifier to cover transient 5xx (500/502/503/504), undici network failures (ECONNRESET/ECONNREFUSED/ETIMEDOUT/UND_ERR_CONNECT_TIMEOUT), and `AbortSignal.timeout`/`AbortController` rejections. Tune `factor: 4`, `retries: 4` so `maxTimeout: 10_000` actually engages on the final attempt (back-off becomes 0.5s/2s/8s/10s ≈ 20.5s across 5 attempts).
- Add `AbortSignal.timeout(EXTERNAL_FETCH_TIMEOUT_MS)` (default 30s, env-overridable) to every external fetch in `fetch.helper.ts` and `copilotAPI.manualFetch`.
- Validate `response.ok` in fetch helpers; non-2xx now throws a typed `HttpFetchError` (status, statusText, url, body — JSON-parsed if possible, raw text otherwise) instead of cryptic JSON-parse failures.
- Remove 15 dead `if (!res) throw new APIError(... 'no response')` guards from `intuitAPI.ts` (97 LOC) now that helpers throw on non-2xx; route real status through `getMessageAndCodeFromError` so `qb_sync_logs` records 503 as 503 instead of 500.
- Update `useSettings`/`useQuickbooks` to wrap `postFetcher` calls in try/catch since helpers now throw.

Worst-case op duration: 5 × 30s + 20.5s ≈ 170s, well within the webhook route's `maxDuration = 300`.

## Test plan

- [x] `yarn vitest run --project unit` (72/72 passing — adds 19 cases covering 5xx retries, undici/cause network errors, TimeoutError, fetch.helper response.ok and timeout paths)
- [x] `yarn lint:check` clean (only pre-existing warnings)
- [x] `yarn prettier:check` clean
- [x] `yarn tsc --noEmit` clean
- [ ] Manual smoke against QBO sandbox: trigger a webhook, point one outbound at an unroutable host to force `ECONNREFUSED`, confirm 5 attempts in `qb_sync_logs` over ~20s before final failure
- [ ] UI spot-check: product/invoice settings save flow still surfaces error state correctly when the API route returns non-2xx

### DIFF
--- a/src/app/api/core/utils/withErrorHandler.ts
+++ b/src/app/api/core/utils/withErrorHandler.ts
@@ -13,7 +13,11 @@ import {
   isIntuitOAuthError,
 } from '@/app/api/core/exceptions/custom'
 import * as Sentry from '@sentry/nextjs'
-import { getMessageAndCodeFromError, RetryableError } from '@/utils/error'
+import {
+  getMessageAndCodeFromError,
+  HttpFetchError,
+  RetryableError,
+} from '@/utils/error'
 import { getCategory } from '@/utils/synclog'
 
 type RequestHandler = (req: NextRequest, params: any) => Promise<NextResponse>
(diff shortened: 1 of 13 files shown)


SandipBajracharya:
@greptileai

SandipBajracharya:
@greptileai

priosshrsth (approved):
lgtm

PR #243, opened 2026-05-07
OUT-3700: remove find-or-map invoice logic from webhook handlers

SandipBajracharya:
## Summary

Surgical revert of [`1ed7b57`](../commit/1ed7b57f18ee8282cf0ce6655ab89c0a620bb96c)'s find-or-map fallback in the live webhook handlers, restoring the original behavior where missing local mapping throws `APIError` so the webhook-level catch records a FAILED row and the resync cron retries once a CREATED resync establishes the mapping.

**Reverted (the find-or-map plumbing):**
- `webhookInvoiceCreated`: dropped early-exit short-circuit
- `webhookInvoicePaid` / `webhookInvoiceVoided`: dropped lazy mapping fallback; now throws `APIError(NOT_FOUND)` on missing mapping
- `handleInvoiceDeleted`: dropped find-or-map call in the "QBO present + no local mapping" branch; now throws `APIError(NOT_FOUND)`
- Removed unused `qbInvoice` param from `findOrMapInvoiceFromQBO`

**Kept intentionally:**
- `checkIfInvoiceExistsInQBO` body unchanged — used by the `syncMissedInvoices` cron path where mapping-on-find is desirable
- `findOrMapInvoiceFromQBO` retained as a private helper (only caller is `checkIfInvoiceExistsInQBO`)
- `handleInvoiceDeleted`'s "QBO absent" branch (soft-delete prior sync logs + record pre-soft-deleted DELETED log)
- OUT-3644 `onConflictDoNothing` race guard and OUT-3655 PENDING-status checks on CREATED log lookups — independent improvements, untouched

**Recovery flow:** throw → `webhook.service.ts` catch transitions PENDING claim → FAILED via `updateOrCreateQBSyncLog` → resync cron retries → succeeds once CREATED resync populates `qb_invoice_sync`.

Linear: https://linear.app/assemblycom/issue/OUT-3700

## Test plan

- [x] Webhook delivery for paid/voided/deleted with no local mapping records a FAILED row (not silent return)
- [x] Resync cron retries those FAILED rows; once a CREATED resync succeeds, subsequent paid/voided/deleted retries find the mapping and resolve to SUCCESS

### DIFF
--- a/src/app/api/quickbooks/invoice/invoice.service.ts
+++ b/src/app/api/quickbooks/invoice/invoice.service.ts
@@ -572,24 +572,6 @@ export class InvoiceService extends BaseService {
     }
 
     const intuitApiService = new IntuitAPI(qbTokenInfo)
-
-    // Check if invoice already exists in QBO (e.g. manually created) and map it if so
-    const mappedFromQBO = await this.findOrMapInvoiceFromQBO({
-      invoiceNumber: invoiceResource.number,
-      copilotInvoiceId: invoiceResource.id,
-      clientId: invoiceResource.clientId,
-      companyId: invoiceResource.companyId,
-      status: invoiceResource.status,
-      total: invoiceResource.total,
-      taxAmount: invoiceResource.taxAmount,
-      intuitApi: intuitApiService,
-    })
-    if (mappedFromQBO) {
-      console.info(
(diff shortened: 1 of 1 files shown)


priosshrsth (approved):
lgtm. But I think we can work on idenfiying pre existing invoice with the invoice number and show error accordingly in another PR.

PR #244, opened 2026-05-08
test(OUT-3708): unit + integration tests for invoice.created webhook

SandipBajracharya:
## Summary

- Adds 9 integration tests + 1 unit test pinning the production behavior of the `invoice.created` webhook handler.
- Mirrors the existing `priceCreated/` integration suite shape (testcontainers + module-mocked Copilot/Intuit).
- Fixes a pre-existing flake in `expiringSweep.test.ts` surfaced by the higher integration-file count.

Linear: [OUT-3708](https://linear.app/assemblycom/issue/OUT-3708)

## What's covered

**Integration (`test/integration/quickbooks/invoiceCreated/`):**
- `happyPath` — OPEN status, mapped product, new customer (asserts mapped-product path via `Description` from `copilot.getProduct`)
- `draftSkip` — DRAFT gate sits before claim
- `idempotency` — pre-existing CREATED log blocks second delivery
- `invoiceAlreadyExists` — `getInvoiceByNumber` short-circuits at the top of `webhookInvoiceCreated`, before customer resolution
- `qbCreateInvoiceFails` — claim row flipped to FAILED with `errorMessage`
- `statusPaidCreatesPayment` — PAID branch creates Payment with `LinkedTxn` linked to the just-created invoice
- `oneOffLineItem` — line item without productId/priceId routes through the Assembly Service ref
- `createNewProductFlagOff` — flag-off + unmapped line item falls back to one-off (no `qb_product_sync` row)
- `useCompanyNameFlag` — company-only payload + flag-on creates customer with `customerType='company'`

**Unit (`test/unit/api/quickbooks/webhook/handleInvoiceCreated.test.ts`):**
- 5 orchestration branches: parse-failure / DRAFT skip / already-claimed / happy path / service-throws → FAILED log

## Side fix

`expiringSweep.test.ts` was using a per-file `vi.mock('@/utils/intuit', …)` factory. Under integration's `pool:forks + fileParallelism:false + isolate:false` config, that mock didn't survive across files once any earlier test transitively loaded the real `@/utils/intuit` (every webhook test does, via `auth.service.ts`). With master's 7 integration files this never landed in the bad slot; with this branch's 16, it landed ~50% of the time.

Fix: moved the module mock to `test/integration/setup.ts` and pinned the singleton on `globalThis` (Vitest re-evaluates the setupFiles factory more than once per session under these flags — a naive shared mock still produces multiple `vi.fn()` instances). Verified with 10 consecutive full-suite runs all green. Full rationale in commit `2664c98` and in the inline comment in `setup.ts`.

## Out of scope (separate tickets)

- INVOICE_UPDATED routing (uses the same handler with `delayMs`)
- absorbedFeeFlag (gates PAYMENT_SUCCEEDED only, not invoice.created)
- Tax math edge cases, fee-paid-by-client line item, sparse-customer-update, excluded-product

## Test plan

- [ ] CI green (`test.yml` + `lint.yml`)
- [ ] Local `yarn test` — 22 files / 101 tests passing (verified 10/10 consecutive runs locally)
- [ ] `yarn lint:check` — clean
- [ ] `yarn prettier:check` — clean
- [ ] priceCreated regression — 6/6 still green
- [ ] Reviewer to spot-check the `globalThis` mock pin in `test/integration/setup.ts` is intelligible


test/helpers/invoiceCreatedTestSetup.ts
```diff
@@ -0,0 +1,43 @@
...
+  beforeEach(async () => {
+    await truncateAllTestTables()
+    const { copilot, intuit } = installMockApis(optsFactory?.())
+    handle.copilot = copilot
+    handle.intuit = intuit
+  })
+
+  afterEach(() => {
+    vi.clearAllMocks()
+  })
```
priosshrsth, line 40: @SandipBajracharya Are we excluding test from tsconfig.json? Ideally we might want the test helpers to be not be imported in our codebase. This is not a blocker. But as we keep adding tests, exporting types and functions will only pollute. Otherwise we might have to block imports from test folder and allow it only from files inside test dir.

test/integration/quickbooks/invoiceCreated/createNewProductFlagOff.test.ts
```diff
@@ -0,0 +1,57 @@
...
+    expect(apis.intuit.getAnItem).toHaveBeenCalledWith('Assembly Service')
+
+    // createItem may run once for the Assembly Service ref (handleServiceItem
+    // → manageServiceItemRef → createItemInQB), but NEVER for the line-item
+    // product. Asserting on the Name field is the strongest signal — '999'
+    // collisions on Id make a count-based check fragile.
+    const createItemNames = apis.intuit.createItem.mock.calls.map(
+      ([payload]) => payload?.Name,
+    )
+    expect(createItemNames).not.toContain('Test Product')
```
priosshrsth, line 34: What is the logic to test for "Test Product" specifically? I though it was  based on json data with id: "2cf93cf0-45fa-485f-b584-03c2c38a3999". If that's the case, should add the name in that json as well. and use that here.

priosshrsth (changes requested):

test/integration/quickbooks/invoiceCreated/createNewProductFlagOff.test.ts
SandipBajracharya: The json file represents how webhook events payload are received from Assembly and they dont include service/product name.  I've added a test data in `test/helpers/mocks.ts:37` file itself.

test/helpers/invoiceCreatedTestSetup.ts
SandipBajracharya: Created separate `tsconfig.json` file for test folder, extending the root one and exclude `test` folder from root `tsconfig.json`

SandipBajracharya:
@greptileai

test/fixtures/invoiceCreated.webhook.json
priosshrsth, line 1: Maybt this can be a .ts file. That will provide additional type safety as well. wdyt?

priosshrsth (approved):
This pr is a  very good start. Well done.

test/fixtures/invoiceCreated.webhook.json
SandipBajracharya: I wanted to make the payload in fixtures as close to what Assembly sends in webhook body. I think your suggestion makes more sense. Thanks

PR #247, opened 2026-05-12
OUT-3543: Type and Zod-validate IntuitAPI responses

SandipBajracharya:
## Summary

Brings every `IntuitAPI` method onto a uniform Zod-parse-at-boundary contract and removes `any` from the IntuitAPI surface. Linear: [OUT-3543](https://linear.app/assemblycom/issue/OUT-3543).

**Before:** ~5 methods parsed; ~13 returned raw responses with no validation. Three `as any` casts, one `Record<string, any>` body, three inline TS interfaces, `_customQuery` typed as `any`.

**After:**
- Every method ends with `Schema.parse(raw)` and declares `Promise<z.infer<typeof Schema>>`.
- `_customQuery: Promise<unknown>` — every caller parses an envelope before reading fields.
- Three hoisted overload types replace `as any` with `as unknown as XxxOverloads`.
- `postFetchWithHeaders` + `postFetcher` body narrowed to `unknown`.
- Row + envelope decomposition applied uniformly across Item / Account / Invoice / Payment / Purchase, with dedicated delete and query envelopes.

```
$ rg ': any|<any>|as any' src/utils/intuitAPI.ts src/type/dto/intuitAPI.dto.ts
(empty)
```

## Out of scope

Documented in `docs/superpowers/specs/2026-05-12-out-3543-intuit-response-types-design.md`. Briefly: `copilotAPI.ts` catch-block narrowing, `withRetry` / `withErrorHandler` generics, axios error guards, framework-boundary `any` (logger / crypto / swr / module decls). Copilot webhook + request-body Zod schemas were already in place pre-branch — the work concentrated on response types where the gap was.

## Design choices

- `.passthrough()` not used; consumers only read modeled fields, default `.strip()` matches intent.
- Fault checks stay imperative (separate from shape parsing).
- `_customQuery` returns `unknown`; callers parse the envelope matching their SQL.
- `CustomerListEnvelopeSchema` wraps `CustomerQueryResponseSchema` directly (no duplicate row schema).

## Follow-ups (not blockers)

- **CI test gap:** integration tests mock `@/utils/intuitAPI` entirely, so production `.parse` never runs in CI. Worth a separate ticket: unit suite parsing captured real QBO sample responses against each schema.
- **`postFetcher` returns `Promise<any>`** — local `invoice` / `item` between fetch and parse is `any`; `.Fault` checks read off `any`. Out of scope but worth tightening next.

## Test plan

- [x] `npx tsc --noEmit` — clean
- [x] `npx vitest run --project unit` — 77/77 pass
- [x] `yarn lint:check` / `yarn prettier:check` — clean
- [x] Full suite at branch HEAD — 87/87 (5 unit + 7 integration via testcontainers)
- [ ] Smoke against QBO sandbox before un-drafting (createItem / createCustomer / createInvoice / createPayment / createPurchase / void / delete)

### DIFF
--- a/src/app/api/quickbooks/product/product.controller.ts
+++ b/src/app/api/quickbooks/product/product.controller.ts
@@ -47,7 +47,6 @@ export async function getItemsFromQB(req: NextRequest) {
   const items = await productService.queryItemsFromQB(
     qbTokenInfo,
     MAX_PRODUCT_LIST_LIMIT,
-    ['Id', 'Name', 'UnitPrice', 'SyncToken', 'Description'],
   )
   return NextResponse.json(items)
 }
--- a/src/app/api/quickbooks/product/product.service.ts
+++ b/src/app/api/quickbooks/product/product.service.ts
@@ -646,13 +646,9 @@ export class ProductService extends BaseService {
     })
   }
 
-  async queryItemsFromQB(
-    qbTokenInfo: IntuitAPITokensType,
-    limit: number,
-    columns: string[],
-  ) {
+  async queryItemsFromQB(qbTokenInfo: IntuitAPITokensType, limit: number) {
     const intuitApi = new IntuitAPI(qbTokenInfo)
-    return await intuitApi.getAllItems(limit, columns)
+    return await intuitApi.getAllItems(limit)
   }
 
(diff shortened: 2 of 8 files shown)


SandipBajracharya:
@greptileai

priosshrsth (approved):

PR #252, opened 2026-05-19
OUT-3754: correct duplicate doc number code from 6240 to 6140

SandipBajracharya:
## Summary

Per [Intuit docs](https://help.developer.intuit.com/s/article/6140-Duplicate-Document-Number-Error), QBO error **6140** is "Duplicate Document Number" and **6240** is "Duplicate Name Exists". `isQBODuplicateDocNumberError` was matching 6240, so the suffix-retry path in `webhookInvoiceCreated` (OUT-3708) never fired on real doc-number collisions — every collision fell through to a FAILED `qb_sync_log`.

Swap 6240 → 6140 in the detector, log/breadcrumb strings, JSDoc, and the docNumber-specific tests. References to 6240 that legitimately refer to "Duplicate Name Exists" (customer DisplayName collision guard) or that just use 6240 as a generic fault-code example are left alone.

## Code sites

- `src/app/api/quickbooks/invoice/invoice.utils.ts` — `isQBODuplicateDocNumberError`
- `src/app/api/quickbooks/invoice/invoice.service.ts` — `resolveAvailableDocNumber` JSDoc + catch/retry log + breadcrumb
- `src/utils/intuitAPI.ts` — `_findInvoicesByDocNumberPrefix` JSDoc reference
- `test/unit/app/api/quickbooks/invoice/invoice.utils.test.ts`
- `test/unit/app/api/quickbooks/invoice/invoice.service.docNumber.test.ts`
- `test/integration/quickbooks/invoiceCreated/qboDocNumberCollision.test.ts`

## Test plan

- [x] `npx vitest run --project unit` — 165/165 passing
- [x] Integration suite (needs Docker) — defer to CI
- [ ] Manual sandbox: trigger a real 6140 in QBO sandbox by enabling "Custom transaction numbers" and reusing a DocNumber; confirm the suffix retry now actually fires (it would not have before this fix)

### DIFF
--- a/src/app/api/quickbooks/invoice/invoice.utils.ts
+++ b/src/app/api/quickbooks/invoice/invoice.utils.ts
@@ -40,7 +40,7 @@ export const findNextAvailableDocNumber = (
 }
 
 /**
- * Recognizes QBO Error 6240 "Duplicate Document Number". Reads `.status`,
+ * Recognizes QBO Error 6140 "Duplicate Document Number". Reads `.status`,
  * `.code`, `errors[].code`, and message text so it works whether the caller
  * surfaces the parsed APIError directly or wraps/normalises it.
  */
@@ -60,16 +60,16 @@ export const isQBODuplicateDocNumberError = (err: unknown): boolean => {
         Detail?: string
         Message?: string
       }
-      if (fault.code === 6240 || fault.code === '6240') return true
+      if (fault.code === 6140 || fault.code === '6140') return true
       if (
-        /6240|Duplicate Document Number/i.test(fault.Detail ?? '') ||
-        /6240|Duplicate Document Number/i.test(fault.Message ?? '')
+        /6140|Duplicate Document Number/i.test(fault.Detail ?? '') ||
+        /6140|Duplicate Document Number/i.test(fault.Message ?? '')
       ) {
         return true
       }
     }
   }
-  if (e.status === 6240 || e.status === '6240') return true
-  if (e.code === 6240 || e.code === '6240') return true
-  return /6240|Duplicate Document Number/i.test(e.message ?? '')
+  if (e.status === 6140 || e.status === '6140') return true
+  if (e.code === 6140 || e.code === '6140') return true
+  return /6140|Duplicate Document Number/i.test(e.message ?? '')
 }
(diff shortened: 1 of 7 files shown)


src/app/api/quickbooks/invoice/invoice.utils.ts
```diff
@@ -60,16 +60,16 @@ export const isQBODuplicateDocNumberError = (err: unknown): boolean => {
         Detail?: string
         Message?: string
       }
-      if (fault.code === 6240 || fault.code === '6240') return true
+      if (fault.code === 6140 || fault.code === '6140') return true
```
priosshrsth:
```suggestion
      if (Number(fault.code) === 6140) return true
```

src/app/api/quickbooks/invoice/invoice.utils.ts
```diff
@@ -60,16 +60,16 @@ export const isQBODuplicateDocNumberError = (err: unknown): boolean => {
...
       ) {
         return true
       }
     }
   }
-  if (e.status === 6240 || e.status === '6240') return true
-  if (e.code === 6240 || e.code === '6240') return true
-  return /6240|Duplicate Document Number/i.test(e.message ?? '')
+  if (e.status === 6140 || e.status === '6140') return true
+  if (e.code === 6140 || e.code === '6140') return true
```
priosshrsth:
```suggestion
  if (Number(e.status) === 6140 || Number(e.code) === 6140) return true
```

priosshrsth (changes requested):
@SandipBajracharya I would suggest to use `QB_STATUS_CODES` constant. So we can do like `QB_STATUS_CODE.DUPLICATE_DOC_NUMBER`, etc. and in comments instead of vague 6140, we say `DUPLICATE_DOC_NUMBER`, wdyt?

SandipBajracharya:
> @SandipBajracharya I would suggest to use `QB_STATUS_CODES` constant. So we can do like `QB_STATUS_CODE.DUPLICATE_DOC_NUMBER`, etc. and in comments instead of vague 6140, we say `DUPLICATE_DOC_NUMBER`, wdyt?

Nice suggestion. Thanks

priosshrsth (approved):
lgtm

PR #253, opened 2026-05-20
OUT-3275: account dropdowns in QB settings

SandipBajracharya:
## Summary

Adds an **Other Settings** accordion section to the QuickBooks integration's settings dashboard that lets portal admins pick which QBO accounts the integration uses:

- **Income account** — default income account assigned to services synced from Assembly.
- **Expense account** — where absorbed invoice payment fees are recorded as expenses.
- **Bank account** — funds those expenses (paired with the expense account).

Linear: https://linear.app/assemblycom/issue/OUT-3275

> **Scope note:** This PR is the dropdown half of OUT-3275 only. The "colon in product name causes duplicate items" half of the same ticket is descoped to a separate effort.

## Architecture

- **Server** — single route `src/app/api/quickbooks/accounts/route.ts` exports both `GET` (returns `{ options, selected }` per bucket) and `PATCH` (validates each ref's `AccountType` against QBO, then updates `qb_portal_connections` scoped by `portalId`).
- **Validation safety** — token fields can't leak: the response uses `QBPortalConnectionSelectSchema.pick({...})` as an allowlist; tests assert `accessToken`/`refreshToken`/`tokenType` are absent from the response.
- **Multi-tenancy** — every read/write scoped via `eq(QBPortalConnection.portalId, this.user.workspaceId)`; tenant-isolation integration test confirms portal A's token cannot mutate portal B's row.
- **QBO query parser quirk** — `_getAccountsForProductMapping` uses three flat queries (one per `AccountType`); `OR`, parentheses, and `IN`-on-`AccountType` all rejected by QBO's parser. Verified empirically in sandbox.
- **UI** — `useOtherSettings` hook gates the GET on `syncFlag && portalConnectionStatus`; when disconnected, shows an inline "Connect to QuickBooks" message and never calls the API. Custom button-based dropdown mirrors the existing QuickBooks-items dropdown styling (no search, no sticky options).

## Commits

```
edbde4c fix(OUT-3275): address pre-push review
1a84634 test(OUT-3275): unit + integration coverage for account-ref endpoints
6b99712 feat(OUT-3275): Other Settings accordion with QBO account dropdowns
1dc37d8 feat(OUT-3275): GET/PATCH /api/quickbooks/accounts and supporting helpers
```

## Test plan

- [x] `yarn test` — 211/211 pass (34 files, unit + integration via testcontainers)
- [x] `npx tsc --noEmit` — zero errors
- [x] `yarn lint:check` + `yarn prettier:check` — clean (10 pre-existing warnings, none new)
- [x] Manual smoke against QBO sandbox — opening "Other Settings" loads all three dropdowns; changing the income account and syncing a Copilot product produces a QBO item with the chosen income account.
- [ ] Reviewer to spot-check the `_getAnAccount` SELECT extension (`AccountType` added) doesn't surprise any existing callers (verified: all 5 existing callers only read `Id`/`Name`/`SyncToken`/`Active`).
- [ ] Reviewer to confirm the QBO `AccountType` narrowing (income subtype = `SalesOfProductIncome`, asset = `Bank`) matches product intent; "user will refine subtypes later" is captured.

## Testing Criteria

https://www.loom.com/share/95f84287758c4917af56ca099e948e9f

## Known follow-ups (separate tickets)

- `checkAndUpdateAccountStatus` self-heal audit — confirm it doesn't silently overwrite a user-chosen ref. Pre-existing concern flagged in the spec, not part of this PR.
- Custom dropdown ARIA: this PR adds `aria-haspopup`/`aria-expanded`/`role="listbox"`/`role="option"`; full keyboard navigation (arrow keys, Escape) deferred.

### DIFF
--- a/src/app/api/quickbooks/accounts/accounts.controller.ts
+++ b/src/app/api/quickbooks/accounts/accounts.controller.ts
@@ -20,22 +20,23 @@ const SafePortalConnectionSchema = QBPortalConnectionSelectSchema.pick({
 export async function listAccounts(req: NextRequest) {
   const user = await authenticate(req)
   const service = new AccountService(user)
-  const data = await service.listAccountsForProductMapping()
-  return NextResponse.json(data)
+  const accountsResponse = await service.listAccountsForProductMapping()
+  return NextResponse.json(accountsResponse)
 }
 
 export async function updateAccountRefs(req: NextRequest) {
   const user = await authenticate(req)
   const body = await req.json()
-  const payload = AccountRefsUpdateSchema.parse(body)
+  const accountRefs = AccountRefsUpdateSchema.parse(body)
 
   const service = new TokenService(user)
-  const conn = await service.updateAccountRefs(payload)
+  const updatedConnection = await service.updateAccountRefs(accountRefs)
 
-  const safe = SafePortalConnectionSchema.parse(conn)
+  const safePortalConnection =
+    SafePortalConnectionSchema.parse(updatedConnection)
 
   return NextResponse.json(
-    { portalConnection: safe },
+    { portalConnection: safePortalConnection },
     { status: httpStatus.OK },
   )
 }
--- a/src/app/api/quickbooks/accounts/accounts.controller.ts
+++ b/src/app/api/quickbooks/accounts/accounts.controller.ts
@@ -1,0 +1,41 @@
+import authenticate from '@/app/api/core/utils/authenticate'
+import { NextRequest, NextResponse } from 'next/server'
+import httpStatus from 'http-status'
+import { AccountService } from '@/app/api/quickbooks/accounts/accounts.service'
+import { TokenService } from '@/app/api/quickbooks/token/token.service'
+import { AccountRefsUpdateSchema } from '@/type/common'
+import { QBPortalConnectionSelectSchema } from '@/db/schema/qbPortalConnections'
+
+// Explicit allowlist of what is safe to return — anything else (especially
+// token/secret fields) is dropped. Adding a new column to QBPortalConnection
+// does NOT widen this surface unless the column is added here too.
+const SafePortalConnectionSchema = QBPortalConnectionSelectSchema.pick({
+  id: true,
+  portalId: true,
+  incomeAccountRef: true,
+  expenseAccountRef: true,
+  assetAccountRef: true,
+})
+
+export async function listAccounts(req: NextRequest) {
+  const user = await authenticate(req)
+  const service = new AccountService(user)
+  const data = await service.listAccountsForProductMapping()
+  return NextResponse.json(data)
+}
+
+export async function updateAccountRefs(req: NextRequest) {
+  const user = await authenticate(req)
+  const body = await req.json()
+  const payload = AccountRefsUpdateSchema.parse(body)
+
+  const service = new TokenService(user)
+  const conn = await service.updateAccountRefs(payload)
+
+  const safe = SafePortalConnectionSchema.parse(conn)
+
+  return NextResponse.json(
+    { portalConnection: safe },
+    { status: httpStatus.OK },
+  )
+}
--- a/src/app/api/quickbooks/accounts/accounts.controller.ts
+++ b/src/app/api/quickbooks/accounts/accounts.controller.ts
@@ -32,11 +32,8 @@ export async function updateAccountRefs(req: NextRequest) {
   const service = new TokenService(user)
   const updatedConnection = await service.updateAccountRefs(accountRefs)
 
-  const safePortalConnection =
-    SafePortalConnectionSchema.parse(updatedConnection)
-
   return NextResponse.json(
-    { portalConnection: safePortalConnection },
+    { portalConnection: SafePortalConnectionSchema.parse(updatedConnection) },
     { status: httpStatus.OK },
   )
 }
--- a/src/app/api/quickbooks/accounts/accounts.service.ts
+++ b/src/app/api/quickbooks/accounts/accounts.service.ts
@@ -1,0 +1,37 @@
+import { BaseService } from '@/app/api/core/services/base.service'
+import { getPortalTokens } from '@/db/service/token.service'
+import APIError from '@/app/api/core/exceptions/api'
+import httpStatus from 'http-status'
+import IntuitAPI from '@/utils/intuitAPI'
+import { AccountsListResponse } from '@/type/common'
+
+export class AccountService extends BaseService {
+  async listAccountsForProductMapping(): Promise<AccountsListResponse> {
+    let tokens
+    try {
+      tokens = await getPortalTokens(this.user.workspaceId)
+    } catch {
+      throw new APIError(
+        httpStatus.NOT_FOUND,
+        'AccountService#listAccountsForProductMapping | no portal connection',
+      )
+    }
+
+    const intuitApi = new IntuitAPI(tokens)
+    const { income, expense, asset } =
+      await intuitApi.getAccountsForProductMapping()
+
+    return {
+      options: {
+        income: income.map((a) => ({ id: a.Id, name: a.Name })),
+        expense: expense.map((a) => ({ id: a.Id, name: a.Name })),
+        asset: asset.map((a) => ({ id: a.Id, name: a.Name })),
+      },
+      selected: {
+        incomeAccountRef: tokens.incomeAccountRef,
+        expenseAccountRef: tokens.expenseAccountRef,
+        assetAccountRef: tokens.assetAccountRef,
+      },
+    }
+  }
+}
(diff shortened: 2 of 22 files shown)


src/app/api/quickbooks/accounts/accounts.controller.ts
```diff
@@ -0,0 +1,41 @@
...
+  return NextResponse.json(data)
+}
+
+export async function updateAccountRefs(req: NextRequest) {
+  const user = await authenticate(req)
+  const body = await req.json()
+  const payload = AccountRefsUpdateSchema.parse(body)
+
+  const service = new TokenService(user)
+  const conn = await service.updateAccountRefs(payload)
```
priosshrsth: Can we use proper variable name to show intention of what it holds?

src/app/api/quickbooks/accounts/accounts.controller.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  const safe = SafePortalConnectionSchema.parse(conn)
+
+  return NextResponse.json(
+    { portalConnection: safe },
+    { status: httpStatus.OK },
```
priosshrsth:
```suggestion
  const portalConnection = await service.updateAccountRefs(payload)

  return NextResponse.json(
    { portalConnection: SafePortalConnectionSchema.parse(portalConnection) },
    { status: httpStatus.OK },
```

src/app/api/quickbooks/accounts/accounts.service.ts
```diff
@@ -0,0 +1,37 @@
...
+  async listAccountsForProductMapping(): Promise<AccountsListResponse> {
+    let tokens
+    try {
+      tokens = await getPortalTokens(this.user.workspaceId)
+    } catch {
+      throw new APIError(
+        httpStatus.NOT_FOUND,
+        'AccountService#listAccountsForProductMapping | no portal connection',
+      )
+    }
```
priosshrsth:
```suggestion
    const tokens = await getPortalTokens(this.user.workspaceId).catch(() => throw new APIError(
        httpStatus.NOT_FOUND,
        'AccountService#listAccountsForProductMapping | no portal connection',
      ));
```

Let's try using `const` if possible.

src/app/api/quickbooks/accounts/accounts.service.ts
```diff
@@ -0,0 +1,37 @@
...
+      options: {
+        income: income.map((a) => ({ id: a.Id, name: a.Name })),
+        expense: expense.map((a) => ({ id: a.Id, name: a.Name })),
+        asset: asset.map((a) => ({ id: a.Id, name: a.Name })),
+      },
+      selected: {
+        incomeAccountRef: tokens.incomeAccountRef,
+        expenseAccountRef: tokens.expenseAccountRef,
+        assetAccountRef: tokens.assetAccountRef,
+      },
```
priosshrsth, line 24: Using incomeAccountId instead of Ref would have added much more clarity imo. But I guess this aligns with the rest of the codebase.

src/app/api/quickbooks/token/token.service.ts
```diff
@@ -90,6 +98,75 @@ export class TokenService extends BaseService {
...
+    let tokens
+    try {
+      tokens = await getPortalTokens(this.user.workspaceId)
+    } catch {
+      throw new APIError(
+        httpStatus.NOT_FOUND,
+        'TokenService#updateAccountRefs | no portal connection',
+      )
+    }
+
```
priosshrsth: Maybe getPotralTokens can throw the error if needed. Or we can have another function `getPortalTokensOrThrow` which'd do the job for us.

src/app/api/quickbooks/token/token.service.ts
```diff
@@ -90,6 +98,75 @@ export class TokenService extends BaseService {
     return token
   }
 
+  async updateAccountRefs(payload: AccountRefsUpdateType) {
+    const parsed = AccountRefsUpdateSchema.parse(payload)
```
priosshrsth:
```suggestion
    const qbAccountRefs = AccountRefsUpdateSchema.parse(payload)
```

src/app/api/quickbooks/token/token.service.ts
```diff
@@ -90,6 +98,75 @@ export class TokenService extends BaseService {
...
+        }
+        if (account.AccountType !== expectedType) {
+          throw new APIError(
+            httpStatus.BAD_REQUEST,
+            `${field} has an incompatible account type`,
+          )
+        }
+      }),
+    )
+
```
priosshrsth: Do we really need all this? Wouldn't zod schema already ensure that `AccountType` is correct?, in which case, can we not just get the object.values() to get ids and then cross check if id is correct?.

Maybe we don't even need to check the id. And if we want to anyway, then yeah I think we can simplify this.

src/app/api/quickbooks/token/token.service.ts
```diff
@@ -90,6 +98,75 @@ export class TokenService extends BaseService {
...
+    // and yields undefined if the WHERE matched nothing. getPortalTokens
+    // above is the de facto guard, but make it explicit so a future refactor
+    // of the ordering doesn't silently produce a misleading 422 from
+    // SafePortalConnectionSchema.parse(undefined) in the controller.
+    if (!conn) {
+      throw new APIError(
+        httpStatus.INTERNAL_SERVER_ERROR,
+        'TokenService#updateAccountRefs | portal connection row not found during update',
+      )
+    }
```
priosshrsth: I think this responsibility should go to `updateQBPortalConnection`, wdyt?. Also maybe the error message is also assuming too much. Could me more generic imo.

src/app/api/quickbooks/token/token.service.ts
```diff
@@ -90,6 +98,75 @@ export class TokenService extends BaseService {
...
+            `${field} has an incompatible account type`,
+          )
+        }
+      }),
+    )
+
+    const conn = await this.updateQBPortalConnection(
+      parsed,
+      eq(QBPortalConnection.portalId, this.user.workspaceId),
+    )
```
priosshrsth, line 104: Why is this named `conn`. I assume this updates our DB, no? `conn` is amigious as if it is an instance of connection with db or quickbook, etc.

Ok I think this aligns with our codebase. So let's keep it as is. I still think at least `portalConnection` would be better. And here I think we can just return without assigning if we throw error on `updateQBPortalConnection`

src/hook/useSettings.ts
```diff
@@ -583,10 +584,111 @@ export const useInvoiceDetailSettings = () => {
...
+    OtherSettingsState | undefined
+  >()
+
+  const { data, error, isLoading } = useSwrHelper(
+    isDisconnected ? null : `/api/quickbooks/accounts?token=${token}`,
+    {
+      suspense: false,
+      revalidateOnMount: true,
+    },
+  )
```
priosshrsth, line 604:
```suggestion
  const { data, error, isLoading } = useSwr<ResponsType>(
    isDisconnected ? null : `/api/quickbooks/accounts?token=${token}`,
    (key) => getFetcher(key)
    {
      revalidateOnFocus: false,
      suspense: false,
      revalidateOnMount: true,
    },
```

You can add global swr config in layout.tsx if needed. Or just add generic support to useSwrHelper.

priosshrsth (changes requested):

src/app/api/quickbooks/token/token.service.ts
SandipBajracharya: This block of code was to cross check if the account ids are actually valid. I guess these are not necessary at this point as those ids are from the intuit itself and since this api is internal, it doesnt make sense to re-verify. Will be removing the whole block. Thanks

src/app/api/quickbooks/token/token.service.ts
SandipBajracharya: keeping it as is. Updated the variable name.

src/app/api/quickbooks/accounts/accounts.service.ts
SandipBajracharya: We are following the same naming pattern for accounts so keeping it as is.

src/hook/useSettings.ts
priosshrsth: @SandipBajracharya I don't think you have addressed this. Currently data is being resolved as any in the codebase.
Screenshot 2026-05-21 at 15 17 33

src/app/api/quickbooks/token/token.service.ts
SandipBajracharya: Updated `getPotralTokens` to throw error from within the function.

src/hook/useSettings.ts
SandipBajracharya: Updated useSwrHelper function to include generic options like `revalidateOnFocus, suspense, revalidateOnMount` and refactored in other call sites.

priosshrsth (approved):
lgtm

PR #256, opened 2026-05-22
OUT-3773: integration tests for payment.succeeded webhook event

SandipBajracharya:
## Summary

- 8 integration tests under `test/integration/quickbooks/paymentSucceeded/` pinning every branch of `WebhookService.handlePaymentSucceeded` — happy path, absorbed-fee flag off, no platform-paid fee, idempotent re-delivery, Copilot invoice not found, local invoice-sync miss, QB `createPurchase` failure, revert-on-log-failure rollback.
- Supporting infra: global `@/utils/sleep` and `afterIfAvailable` mocks in the integration setup, two new seed constants, expanded `createMockIntuitAPI` defaults (`createPurchase` / `deletePurchase`), id-echoing `getAnAccount`, real `getInvoice` default, new `setupPaymentSucceededTest` helper, new webhook fixture.

## Production finding (separate follow-up worth filing)

`webhook.service.ts:518` throws `APIError(404)` **outside** the try/catch on line 524. When Copilot returns no invoice, the handler responds 404 and leaves a PENDING claim row instead of writing FAILED. The `copilotInvoiceNotFound` test pins this current behavior; the fix is to move the check inside the try.

## Test plan

- [x] `npx vitest run --project integration test/integration/quickbooks/paymentSucceeded/` → 8 files / 9 tests pass
- [x] `npx vitest run --project integration` → 29 files / 40 tests pass (no regressions)
- [x] `npx vitest run --project unit` → 15 files / 238 tests pass
- [x] `yarn lint:check` clean (0 errors, pre-existing useEffect warnings only)
- [x] `yarn prettier:check` clean
- [x] CI Lint workflow green on push

Linear: https://linear.app/assemblycom/issue/OUT-3773

### DIFF
--- a/test/fixtures/paymentSucceeded.webhook.ts
+++ b/test/fixtures/paymentSucceeded.webhook.ts
@@ -15,7 +15,7 @@ type Envelope = {
 type PaymentSucceededFixture = Envelope &
   z.input<typeof PaymentSucceededResponseSchema>
 
-const paymentSucceededPayload: PaymentSucceededFixture = {
+export const paymentSucceededPayload: PaymentSucceededFixture = {
   eventType: 'payment.succeeded',
   object: 'payment',
   data: {
@@ -28,5 +28,3 @@ const paymentSucceededPayload: PaymentSucceededFixture = {
     createdAt: '2024-02-21T15:31:16.789Z',
   },
 }
-
-export default paymentSucceededPayload
--- a/test/fixtures/paymentSucceeded.webhook.ts
+++ b/test/fixtures/paymentSucceeded.webhook.ts
@@ -1,0 +1,32 @@
+import type { z } from 'zod'
+
+import { PaymentStatus } from '@/app/api/core/types/invoice'
+import { PaymentSucceededResponseSchema } from '@/type/dto/webhook.dto'
+import {
+  TEST_COPILOT_INVOICE_ID,
+  TEST_COPILOT_PAYMENT_ID,
+} from '@test/helpers/seed'
+
+type Envelope = {
+  eventType: 'payment.succeeded'
+  object: 'payment'
+}
+
+type PaymentSucceededFixture = Envelope &
+  z.input<typeof PaymentSucceededResponseSchema>
+
+const paymentSucceededPayload: PaymentSucceededFixture = {
+  eventType: 'payment.succeeded',
+  object: 'payment',
+  data: {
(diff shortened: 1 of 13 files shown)


SandipBajracharya:
@greptileai

test/fixtures/paymentSucceeded.webhook.ts
```diff
@@ -0,0 +1,32 @@
...
+    invoiceId: TEST_COPILOT_INVOICE_ID,
+    status: PaymentStatus.SUCCEEDED,
+    paymentMethod: 'creditCard',
+    brand: 'visa',
+    feeAmount: { paidByPlatform: 2500, paidByClient: 0 },
+    createdAt: '2024-02-21T15:31:16.789Z',
+  },
+}
+
+export default paymentSucceededPayload
```
priosshrsth: @SandipBajracharya This might be opinionated. But named exports are better than default export. I adds consistency. 

eg. paymentSucceededPayload can be used like anything now. But with named export we can ensure that it is used with the same name everywhere. This will be a huge help in readability, and refactoring if needed.

priosshrsth (changes requested):

test/integration/quickbooks/paymentSucceeded/revertOnLogFailure.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    // outer catch's FAILED log falls through to the real impl. try/finally
+    // restores the spy — leaked spies contaminate later tests under
+    // isolate:false.
+    const spy = vi
+      .spyOn(SyncLogService.prototype, 'updateOrCreateQBSyncLog')
+      .mockImplementationOnce(async () => {
+        throw new Error('Sync log write failed')
+      })
+
+    try {
```
priosshrsth: Interesting. Do we need try catch?
Can you try doing it without try catch?

And doing mockRestore with afterEach instead?

My assumption is even if it fails, it should not crash the test.

priosshrsth (changes requested):

test/integration/quickbooks/paymentSucceeded/revertOnLogFailure.test.ts
SandipBajracharya: There is no try/catch but try/finally. Your approach is cleaner so implementing that. Thanks!

priosshrsth (approved):
lgtm

PR #257, opened 2026-06-03
OUT-3814: one item per product, driven by product.created

SandipBajracharya:
## What & why

Part of **OUT-3787** (simplified product mapping). Previously each Assembly product *price* created a separate QB Item with auto-suffixed names ("Name (1)", "(2)"), cluttering the QB item list. This maps **one Assembly product → one QB Item**, with invoice lines carrying their own price.

Scope of this PR: **backend + tests only.** The mapping-UI changes ship in a follow-up PR (OUT-3816); the DB migration/index ships in OUT-3815.

## Changes

- **One item per product.** Mapping lookup is product-keyed (`getMappingByProductId`, earliest-created canonical). `getOne` and `getAllByProductId` filter `deletedAt`.
- **Item creation driven by `product.created`** (`webhookProductCreated`) instead of `price.created`; the `price.created` event/handler is removed. Items are created at `UnitPrice: 0`.
- **Invoice line `UnitPrice` comes from the line amount** (`lineItem.amount / 100`); removed `handleItemAmount` and the `copilotUnitPrice` round-trip. Net: no invoice amount changes (lines already passed line-level price).
- **Lazy invoice-time item creation** aligned: no price fetch, one un-suffixed item per product, product-only upsert.
- `handleProductMap` upserts per product; `ProductChangedItemReferenceSchema` drops `priceId`/`numericPrice`.
- Resync (`sync.service.processProductCreate`) reconstructs from the product.

## Tests

- `price.created` integration suite reworked into `product.created` (happy path, flag-off, idempotency, QB failure, soft-deleted-mapping).
- New: invoice line drives `UnitPrice`; lazy item creation makes one un-suffixed item; `product/map` product-keyed upsert.
- Full suite green (286), `tsc` clean, lint/prettier clean.

## Testing Criteria

https://www.loom.com/share/c01657c34e08491ea67ff25e42f64756

## Notes for review

- `qb_product_sync` price columns (`price_id`, `unit_price`, `copilot_unit_price`) are intentionally left in the schema; old rows keep their data, new rows leave them null. The partial unique index + one-time cleanup land in **OUT-3815**.
- Follow-up **OUT-3816** carries the flatten DTO narrowing + `getProductsForMapping` price removal + the settings-table UI.


SandipBajracharya:
Re: P2 — logSync CREATED upsert key not unique in the DB

Thanks — verified, with one correction to the mechanism:

`updateOrCreateQBSyncLog` doesn't use an unordered `findFirst`. It goes through `SyncLogService.getOne`, which orders by `createdAt` (default `desc`):

```ts
async getOne(conditions, orderByDirection: 'asc' | 'desc' = 'desc') {
  return this.db.query.QBSyncLog.findFirst({
    where: conditions,
    orderBy: [orderMap[orderByDirection](QBSyncLog.createdAt)],
  })
}
```

So when multiple `PRODUCT/CREATED` rows match `(portalId, copilotId, eventType)`, it deterministically updates the **most recent** one — not a random row, so no row goes stale non-deterministically.

The residual point is fair: multiple `PRODUCT/CREATED` rows *can* coexist (legacy data, where old logs were keyed per `copilotPriceId`, or the `webhookProductCreated` ↔ lazy-creation race), and only the newest is reconciled. Impact is low, though — a new product produces exactly one CREATED log; legacy products don't re-fire `product.created`, so their existing rows aren't touched; and this is an audit/retry-log artifact, not QB-facing data.

One correction on the mitigation: **OUT-3815 is the `qb_product_sync` index/cleanup and does not touch `qb_sync_logs`**, so the CREATED-log dedup isn't actually covered there. The existing `uq_qb_sync_logs_oneshot_active` partial index covers invoice/payment one-shots, not `PRODUCT/CREATED`.

Treating this as non-blocking given the determinism above and the low impact. If we want DB-enforced dedup, I'll file a separate ticket for a partial unique index on `qb_sync_logs (portal_id, copilot_id, event_type) WHERE entity_type = 'product' AND event_type = 'created'`.

src/app/api/quickbooks/invoice/invoice.service.ts
```diff
@@ -287,7 +225,7 @@ export class InvoiceService extends BaseService {
       qbItem = await productService.createItemInQB(
         {
           productName: newName,
-          unitPrice: priceInfo.amount,
+          unitPrice: 0,
```
priosshrsth: I think this is just telling a lie again. Maybe this should be nullable instead? Or maybe have a default product price, if there is any. What do you think. Does it make sense to be nullable?

src/app/api/quickbooks/product/product.service.ts
```diff
@@ -536,62 +542,39 @@ export class ProductService extends BaseService {
...
+  async webhookProductCreated(
+    resource: ProductCreatedResponseType,
     qbTokenInfo: IntuitAPITokensType,
   ): Promise<void> {
-    const priceResource = resource.data
-    addSyncBreadcrumb('Price created flow started', {
-      priceId: priceResource.id,
-      productId: priceResource.productId,
+    const productResource = resource.data
+    addSyncBreadcrumb('Product created flow started', {
```
priosshrsth:
```suggestion
    addSyncBreadcrumb('Product creation flow started', {
```

src/app/api/quickbooks/product/product.service.ts
```diff
@@ -536,62 +542,39 @@ export class ProductService extends BaseService {
...
-
-      // total products with the same product id
-      const itemsCount = mappedProducts?.length
-
-      // filter out with priceId
-      const productWithPriceCount = mappedProducts?.filter(
-        (product) => product.priceId === priceResource.id,
-      ).length
+      // 01. if this product is already mapped to a QB item, do nothing.
+      const mappedProducts = await this.getAllByProductId(
```
priosshrsth: It looks like we only care about the count here. So maybe only query count? count queries are generally faster as well. wdyt?

Or maybe even an existence like find first id with the query..

src/app/api/quickbooks/product/product.service.ts
```diff
@@ -536,62 +542,39 @@ export class ProductService extends BaseService {
...
+        ['id'],
+      )
 
       addSyncBreadcrumb('Product mapping check', {
-        alreadyMapped: !!productWithPriceCount,
+        alreadyMapped: !!mappedProducts?.length,
       })
-      if (productWithPriceCount && productWithPriceCount > 0) {
-        console.info('Product already mapped with price')
+      if (mappedProducts && mappedProducts.length > 0) {
```
priosshrsth:
```suggestion
      if (mappedProducts?.length) {
```

src/app/api/quickbooks/product/product.service.ts
```diff
@@ -604,43 +587,37 @@ export class ProductService extends BaseService {
...
           qbTokenInfo.incomeAccountRef,
         )
-        // create item in QB
+        // create item in QB. No price at product.created time — invoice lines
+        // carry their own UnitPrice, so the item's catalog price is 0.
         qbItem = await this.createItemInQB(
           {
             productName: z.string().parse(qbItemName),
-            unitPrice: priceResource.amount,
+            unitPrice: 0,
```
priosshrsth: Again, I am not aligned with setting unitPrice to 0. Maybe this makes sense. But I do need more context from you.

src/app/api/quickbooks/webhook/webhook.service.ts
```diff
@@ -409,35 +409,34 @@ export class WebhookService extends BaseService {
...
       console.error(
-        'WebhookService#handlePriceCreated | Could not parse price created resource',
+        'WebhookService#handleProductCreated | Could not parse product created resource',
       )
       return
     }
-    const parsedCreatedPriceResource = parsedCreatedPrice.data
-    const priceResource = parsedCreatedPriceResource.data
+    const parsedCreatedProductResource = parsedCreatedProduct.data
+    const productResource = parsedCreatedProductResource.data
```
priosshrsth, line 426:
```suggestion
    const { data: parsedCreatedProductResource } = ProductCreatedResponseSchema.safeParse(payload)
    if (!parsedCreatedProductResource) {
console.error(
        'WebhookService#handleProductCreated | Could not parse product created resource',
)
return
}
    const productResource = parsedCreatedProductResource.data

```

I think this can be simplified like this.

priosshrsth (changes requested):
@SandipBajracharya The PR looks good to me. I have added some comments and suggestions. But all of them are opinionated and minor. Let me know what you think about them

src/app/api/quickbooks/webhook/webhook.service.ts
SandipBajracharya: We cannot access data attribute unless we know the parse is successful. so this line `const { data: parsedCreatedProductResource } = ProductCreatedResponseSchema.safeParse(payload)` wont be valid.

src/app/api/quickbooks/invoice/invoice.service.ts
SandipBajracharya: Actually did quick research and found that unitPrice field is not required to create item in QBO. Removing this row.

priosshrsth (approved):
lgtm

PR #258, opened 2026-06-03
OUT-3816: collapse product mapping table to one row per product

SandipBajracharya:
## What & why

Part of **OUT-3787** (simplified product mapping). The mapping settings table previously rendered one row per *price*; this collapses it to **one row per Assembly product** and removes price from the table entirely (per product → one QB item; invoice lines carry their own price).

> **Stacked on #257 (OUT-3814).** Base is `OUT-3814`; retarget to `master` once #257 merges.

## Changes

- **Table is one row per product**, matched by `productId` (dropped `priceId` predicates in `useMapItem` / `useProductTableSetting` / `selectItem`).
- **No prices in the table** — removed the Assembly product price and the QB item price (subtitle + dropdown).
- **`getProductsForMapping`** (was `getProductsWithPrices`) returns one row per product and no longer fetches Copilot prices; flatten DTO is name-only.
- **Dropped price validations** from the mapping save: removed the `unitPrice is required when isExcluded is false` superRefine and `numericPrice` from `QBItemSchema`.
- **`unitPrice` is now always `null`** through the UI mapping path and `handleProductMap`; removed dead `price`/`numericPrice` from `QBItemDataType` / `formatQBItemForListing`.

## Regression review fixes (last commit)

- `webhookProductUpdated` now sends `sparse: true` so a name/description update doesn't reset a mapped QB item's `UnitPrice` to `0` (mapped items store `null` unitPrice now).
- `handleProductMap` calls `unsetTransaction()` before the initial-save early return (no closed-tx singleton leak).
- Removed stale `numericPrice` from a test fixture; dropped the unused `createdAt` from the flatten DTO.

## Verification

- `tsc` clean · full suite **286 passing** · lint/prettier clean.
- Manual smoke test (dashboard: multi-price product shows one row; select/exclude persists after Save) still **pending** — needs a portal with a multi-price product.


SandipBajracharya:
@greptileai

src/app/api/quickbooks/product/product.service.ts
```diff
@@ -323,78 +322,28 @@ export class ProductService extends BaseService {
...
-          list.push(price)
-        } else {
-          grouped.set(price.productId, [price])
-        }
-      }
-      nextToken = page.nextToken
-    } while (nextToken)
+  async getProductsForMapping(): Promise<ProductFlattenArrayResponseType> {
+    const copilot = new CopilotAPI(this.user.token)
+    const products = await copilot.getProducts(
```
priosshrsth: Minor nitpick. If getProducts is our implementation than can we fix this params? Maybe it can take an object. Rule of thumb is more than 2 params, object params is better rather than position params.

src/app/api/quickbooks/product/product.service.ts
```diff
@@ -323,78 +322,28 @@ export class ProductService extends BaseService {
...
+    )
 
-    return grouped
+    const formatted = (products?.data ?? []).map((product) => ({
+      id: product.id,
+      name: product.name,
+      description: convert(product.description),
+    }))
+
+    return { products: formatted }
```
priosshrsth, line 341:
```suggestion
    reeturn (products?.data ?? []).map((product) => ({
      id: product.id,
      name: product.name,
      description: convert(product.description),
    }))
```

We are returning products here. No need to make it an object imo.

src/hook/useSettings.ts
```diff
@@ -194,11 +188,7 @@ export const useProductMappingSettings = () => {
       [index]: '',
     }))
     const fileteredChangedItem = changedItemReference.filter(
-      (item) =>
-        !(
-          item.id === products[index].id &&
-          item.priceId === products[index].priceId
-        ),
+      (item) => item.id !== products[index].id,
```
priosshrsth: This can potentiall lead to error. products[index] could be undefined.
```suggestion
      (item) => item.id !== products[index]?.id,
```

src/hook/useSettings.ts
```diff
@@ -210,18 +200,13 @@ export const useProductMappingSettings = () => {
 
     // update the mapped array
     const mappedArray = mappingItems.map((mapItem) => {
-      if (
-        mapItem.productId === products[index].id &&
-        mapItem.priceId === products[index].priceId
-      ) {
+      if (mapItem.productId === products[index].id) {
```
priosshrsth: Doesn't this line show type error? I think Array<number> should extend undefined as well mostly unless it is a tuple.

priosshrsth (changes requested):

src/app/api/quickbooks/product/product.service.ts
SandipBajracharya: Letting this be for now.

src/hook/useSettings.ts
SandipBajracharya: No error in compile time. This can surely throw error in runtime. Added optional chaining.

src/app/api/quickbooks/product/product.service.ts
priosshrsth: Ok

priosshrsth (approved):
lgtm

PR #263, opened 2026-06-29
OUT-3867: finalize non-open invoice voids instead of looping retries

SandipBajracharya:
## Problem (OUT-3867)

Sentry paged `SyncService#intiateSync | Records exceeded max retry count` for an `invoice/voided` log with `errorCategory: others` and `errorMessage: null`.

Root cause: when a void webhook arrived for an invoice whose mapping was **not** `OPEN`, `webhookInvoiceVoided` did a bare `return` — no terminal log write. That left the `qb_sync_logs` claim stuck in `PENDING`. The stale-claim reaper (`flipStalePendingToFailed`) later promoted it to `FAILED`/`OTHERS` with no message (a heuristic, not a captured error). Since `OTHERS` is retryable and a non-OPEN invoice never returns to `OPEN`, every resync re-hit the same bare `return`, incrementing `attempt` until it hit `MAX_ATTEMPTS = 25` and paged.

Two symptoms: (1) null `errorMessage` column, (2) 25-retry storm + false alert.

## Fix

The non-OPEN branch now finalizes the log terminally instead of bare-returning:

- **already `VOID`** → idempotent SUCCESS via `logSync` (goal state already reached; no re-void)
- **any other non-OPEN** (PAID/DELETED/…) → non-retryable `FAILED` (`shouldRetry: false`, `category: VALIDATION`) with a clear `errorMessage` and **no** `errorCode`, so the IU failure notifier stays a no-op

`shouldRetry: false` keeps the resync cron from ever re-selecting the row, so no retry loop and no max-retry page.

Note: the `handleInvoiceDeleted` non-voided branch intentionally stays retryable — a delete can arrive while the invoice is still `OPEN`, and the companion void event flips it to `VOID` so a retry later succeeds (inverse recovery dynamics vs. void).

## Tests

- `nonOpenStatus.test.ts` rewritten (previously asserted the buggy `PENDING` outcome) → now asserts non-retryable `FAILED` with a message, nothing voided.
- `alreadyVoided.test.ts` (new) → already-VOID mapping yields SUCCESS with no re-void.

Verification: voided suite 9 passed · full integration 72 · unit 246 · eslint/prettier/tsc clean.

### DIFF
--- a/src/app/api/quickbooks/invoice/invoice.service.ts
+++ b/src/app/api/quickbooks/invoice/invoice.service.ts
@@ -2,7 +2,12 @@ import APIError from '@/app/api/core/exceptions/api'
 import User from '@/app/api/core/models/User.model'
 import { BaseService } from '@/app/api/core/services/base.service'
 import { InvoiceStatus, SyncableEntity } from '@/app/api/core/types/invoice'
-import { EntityType, EventType, LogStatus } from '@/app/api/core/types/log'
+import {
+  EntityType,
+  EventType,
+  FailedRecordCategoryType,
+  LogStatus,
+} from '@/app/api/core/types/log'
 import { CustomerService } from '@/app/api/quickbooks/customer/customer.service'
 import {
   findNextAvailableDocNumber,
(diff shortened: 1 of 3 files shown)


SandipBajracharya:
@greptileai

priosshrsth (approved):
lgtm

PR #264, opened 2026-07-08
OUT-3874: use invoice address to designate tax jurisdiction

SandipBajracharya:
## What & why

[OUT-3874](https://linear.app/assemblycom/issue/OUT-3874/use-address-to-designate-tax-jurisdiction) — attach the customer's address to the QuickBooks invoice so the tax jurisdiction is visible on the QBO record for reporting/review.

## Behavior

- **Tax total is always sourced from Assembly.** `TxnTaxDetail.TotalTax` is set to Assembly's computed tax on every invoice; the sync log and any created payment use Assembly's local total. QBO is **not** relied on to compute tax.
- **Customer address is sent for the QBO record**, as both `BillAddr` and `ShipAddr`, only when the two fields QBO needs for a jurisdiction — state and postal code — are present. `getUsStateCode` (via `us-state-converter`) maps a full state name to the two-letter `CountrySubDivisionCode`.
- Address sub-fields are optional (shared `AddressSchema`) so a partial address on the REST fetch paths (`_getInvoice`/`_getInvoices`, which use `.parse()`) can't abort a bulk resync.
- **Refactor**: consolidated the duplicate `InvoiceDeleted`/`InvoiceVoided` schemas into `InvoiceDestructiveResponseSchema`.

## Testing

- Full suite green (`yarn test`); `tsc` and lint clean.
- Unit tests for `getUsStateCode`; integration coverage for the address payload (`BillAddr`/`ShipAddr`), the always-sent `TxnTaxDetail`, the missing-postal-code skip, and the paid-invoice payment total.

## Open item (needs sandbox confirmation)

On realms with **Automated Sales Tax enabled**, sending `ShipAddr` may cause QBO to recompute tax from that jurisdiction and ignore the `TxnTaxDetail.TotalTax` we send — which would contradict the "address is report-only" intent. This should be verified against a QBO sandbox before relying on it; if AST overrides, we should send `BillAddr` only (lower risk) or accept QBO-computed tax deliberately. The integration tests validate the payload we send, not QBO's tax outcome.

### DIFF
--- a/src/app/api/quickbooks/invoice/invoice.service.ts
+++ b/src/app/api/quickbooks/invoice/invoice.service.ts
@@ -729,17 +729,15 @@ export class InvoiceService extends BaseService {
     const addressPayload =
       invoiceResource.address?.postalCode && countrySubDivisionCode
         ? {
-            Line1: invoiceResource.address.addressLine1,
+            Line1:
+              invoiceResource.address.addressLine1 ||
+              invoiceResource.address.addressLine2,
             City: invoiceResource.address.city,
             CountrySubDivisionCode: countrySubDivisionCode,
             PostalCode: invoiceResource.address.postalCode,
             Country: invoiceResource.address.country,
           }
         : null
-
-    // TODO: tax-exempt invoices also have totalTax === 0 and currently fall
-    // into QBO-computes; gate on an explicit exempt flag when that's needed.
-    const overrideTax = totalTax > 0 || !addressPayload
 
     const buildPayload = (resolvedDocNumber: string) => ({
       Line: lineItems,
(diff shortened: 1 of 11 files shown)


SandipBajracharya:
@greptileai re-review whole pr changes again

src/app/api/quickbooks/invoice/invoice.service.ts
```diff
@@ -720,6 +721,24 @@ export class InvoiceService extends BaseService {
     const billEmailAddress =
       customer?.PrimaryEmailAddr?.Address || existingCustomer?.email
 
+    const countrySubDivisionCode = getUsStateCode(
+      invoiceResource.address?.region,
+    )
+
+    // QBO needs both state and postal code to resolve a tax jurisdiction.
+    const addressPayload =
+      invoiceResource.address?.postalCode && countrySubDivisionCode
```
priosshrsth, line 730: This is a minor nitpick and no action needed here. But maybe switching `countrySubDivisionCode` first might be better.

src/type/dto/webhook.dto.ts
```diff
@@ -24,6 +24,17 @@ export const InvoiceLineItemSchema = z.object({
...
+// a partial address doesn't throw on the REST .parse() paths.
+export const AddressSchema = z.object({
+  addressLine1: z.string().optional(),
+  addressLine2: z.string().optional(),
+  city: z.string().optional(),
+  country: z.string().optional(),
+  region: z.string().optional(),
+  postalCode: z.string().optional(),
+})
+
```
priosshrsth, line 37: @SandipBajracharya I would like to suggest using .nullish() instead of optional() here. In case we can have city with null value and such.

priosshrsth (approved):
@SandipBajracharya Minor nitpick. But pr looks good to merge.

src/type/dto/webhook.dto.ts
SandipBajracharya: I confirmed that either the value is there or the key wont appear in the payload

PR #266, opened 2026-07-22
OUT-3604: Stripe payout → batched QBO bank deposit

SandipBajracharya:
## What

On `payout.reconciliation_completed`, create **one QBO Bank Deposit per Stripe payout** whose net matches the bank-feed entry, so QBO auto-suggests a one-click reconciliation match. Replaces the never-shipped per-payment deposit path.

Assembly listens to Stripe's `payout.reconciliation_completed`, resolves each charge to a Copilot invoice, and forwards an enriched webhook. This app resolves each invoice → QBO Payment via `qb_sync_logs`, asserts `sum(gross) − sum(fee) == payout.netAmount` (in cents), and creates one Bank Deposit: N payment-linked lines + one negative aggregated-fee line.

## Commits
- **apply migrations one-per-transaction (`db:migrate`)** — drizzle batches all pending migrations into one transaction, which breaks enum-add-then-use; `migratePerFile` applies one file per transaction (wired into `build.sh`).
- **payout/settled enums, bank-deposit columns, idempotency index** — `PAYOUT`/`SETTLED`, the `bank_deposit_fee_flag` + `bank_account_ref` columns (schema + migration together), the one-shot unique index + `claimWebhookEvent` predicate extended for `payout/settled`, `getSuccessfulPaidPaymentIds`.
- **type the QBO createDeposit response** — `QBDepositResponseSchema` + `assertNotQBFault`/Zod-parse.
- **resolve the bank account ref** — `AccountTypeObj.Bank` (reactivate archived, throw on deleted), `bankAccountRef` threaded through all token-construction sites.
- **create one batched bank deposit** — the payout handler, batched deposit builder, `payment.succeeded` no-op in batched mode, `invoice.paid` → Undeposited Funds routing, resync skip for payouts.

## Behavior (`bankDepositFeeFlag` on)
| Event | Action |
| --- | --- |
| `invoice.paid` | Payment → Undeposited Funds |
| `payment.succeeded` | no-op (no per-payment expense/deposit) |
| `payout.reconciliation_completed` | one batched Bank Deposit (net of fees) |

Aborts with a FAILED log (never a partial/unbalanced deposit) on: refund lines, negative aggregate fee, duplicate/unresolved invoices, or a sum mismatch.

## Notes / follow-ups
- Test files (integration suite + fixtures + the `captureWebhookEventGET` harness) are managed separately and are **not** in these commits.
[...]

PR #267, opened 2026-07-23
OUT-4003: bank deposit settings — dedicated toggle + bank account dropdown

SandipBajracharya:
Stacked on #266 (OUT-3604). Base: `OUT-3604-new-re`.

## What
Adds the settings-dashboard UI + API so a user can enable batched Stripe→QBO bank deposits and choose the deposit bank account. Without this, `bankAccountRef` is never set and the payout handler aborts with "bank account not configured."

## Commits
- **schema** — `SettingRequestSchema` + `InvoiceSettingType` gain `bankDepositFeeFlag` + `bankAccountRef`; invoice type requires the flag, and requires a non-empty `bankAccountRef` when the flag is on (surfaces as 422).
- **endpoint** — `GET /api/quickbooks/setting/bank-account` (route → controller → `BankAccountService`) lists active QBO Bank accounts; selects the full account column set so it parses against real QBO, `maxresults 100`.
- **persistence** — `setting.controller` GET returns `bankAccountRef`; POST persists it transactionally with the invoice flags (empty → null; `setTransaction`/`unsetTransaction` paired for both services).
- **UI** — extract shared `AccountSelect`; independent `bankDepositFeeFlag` toggle + "Deposit bank account" dropdown in `InvoiceDetail`; inline hint/error; `canSave` disables the Invoice Update button until an account is chosen.

## Verification
- `tsc` 0 errors, lint clean, `next build` OK, full suite 92 files / 342 tests pass.
- Manual UI verification done.
- Full-stack reviewed; findings fixed (surface bank-account fetch errors, extract the service layer, `maxresults` cap, empty→null coercion).

https://www.loom.com/share/243d1bee5d7b446a8c29e3619c4cc692

## Notes
- Test files (schema unit test + `setting` integration tests) and their test-infra are managed separately and are **not** in these commits — CI on this branch won't exercise the new tests until that lands.
- Deferred follow-ups: disambiguate the two "bank account" labels ("Deposit bank account" vs the asset "Bank account" in Account Mapping); collapse the `getSettings` double DB round-trip.

### DIFF
--- a/src/app/api/quickbooks/setting/bank-account/bank-account.service.ts
+++ b/src/app/api/quickbooks/setting/bank-account/bank-account.service.ts
@@ -1,0 +1,12 @@
+import { BaseService } from '@/app/api/core/services/base.service'
+import { QBAccountQueryResponseSchema } from '@/type/dto/intuitAPI.dto'
+import IntuitAPI, { QB_ACCOUNT_COLUMNS } from '@/utils/intuitAPI'
+
+export class BankAccountService extends BaseService {
+  async listActiveBankAccounts(intuitApi: IntuitAPI) {
+    const rawResult = await intuitApi.customQuery(
+      `SELECT ${QB_ACCOUNT_COLUMNS.join(', ')} FROM Account WHERE AccountType = 'Bank' AND Active = true maxresults 100`,
+    )
+    return QBAccountQueryResponseSchema.parse(rawResult ?? {}).Account ?? []
+  }
+}
--- a/src/app/api/quickbooks/setting/bank-account/bank-account.service.ts
+++ b/src/app/api/quickbooks/setting/bank-account/bank-account.service.ts
@@ -4,8 +4,10 @@ import IntuitAPI, { QB_ACCOUNT_COLUMNS } from '@/utils/intuitAPI'
 
 export class BankAccountService extends BaseService {
   async listActiveBankAccounts(intuitApi: IntuitAPI) {
+    // 1000 is QBO's max single-page size — far more bank accounts than any
+    // real company has, so a single query returns them all.
     const rawResult = await intuitApi.customQuery(
-      `SELECT ${QB_ACCOUNT_COLUMNS.join(', ')} FROM Account WHERE AccountType = 'Bank' AND Active = true maxresults 100`,
+      `SELECT ${QB_ACCOUNT_COLUMNS.join(', ')} FROM Account WHERE AccountType = 'Bank' AND Active = true maxresults 1000`,
     )
     return QBAccountQueryResponseSchema.parse(rawResult ?? {}).Account ?? []
   }
--- a/src/app/api/quickbooks/setting/setting.controller.ts
+++ b/src/app/api/quickbooks/setting/setting.controller.ts
@@ -36,11 +36,10 @@ export async function getSettings(req: NextRequest) {
   }
   const setting = await settingService.getOneByPortalId(returningFields)
 
-  let bankAccountRef: string | null = null
-  if (parsedType.success && parsedType.data === SettingType.INVOICE) {
-    const portalConnection = await getPortalConnection(user.workspaceId)
-    bankAccountRef = portalConnection?.bankAccountRef || null
-  }
+  const bankAccountRef =
+    parsedType.success && parsedType.data === SettingType.INVOICE
+      ? (await getPortalConnection(user.workspaceId))?.bankAccountRef || null
+      : null
 
   return NextResponse.json({ setting, bankAccountRef })
 }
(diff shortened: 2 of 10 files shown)


src/components/dashboard/settings/sections/account/AccountSelect.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const dropdownRef = useRef<HTMLDivElement>(null)
+  const buttonRef = useRef<HTMLButtonElement>(null)
+
+  useClickOutside(dropdownRef, () => setIsOpen(false), [buttonRef])
+
+  const disabled = !options || options.length === 0
+  const selected = options?.find((o) => o.id === value)
+  // Defends against an account being deleted in QBO between load and save —
+  // surfaces the stale id with a hint instead of silently snapping to empty.
+  const optionMissing = !!value && !selected
```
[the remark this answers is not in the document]
SandipBajracharya: Fixed in e0b1af2 — AccountSelect now shows 'Loading accounts…' while the fetch is in flight (options === undefined), distinct from 'No matching accounts' (options.length === 0), and InvoiceDetail only shows the 'select an account' hint once options have loaded.

src/hook/useSettings.ts
```diff
@@ -448,16 +453,31 @@ export const useInvoiceDetailSettings = () => {
...
     flag: keyof InvoiceSettingType,
-    state: boolean,
+    value: boolean | string,
   ) => {
-    setSettingState((prev) => ({
-      ...prev,
-      [flag]: state,
-    }))
+    setSettingState((prev) => ({ ...prev, [flag]: value }))
   }
```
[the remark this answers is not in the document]
SandipBajracharya: Fixed in e0b1af2 — changeSettings is now generic over the field key: `<K extends keyof InvoiceSettingType>(flag: K, value: InvoiceSettingType[K])`, so the value type is tied to the key and a string for a boolean flag no longer type-checks.

SandipBajracharya:
@greptileai review whole thing again

src/app/api/quickbooks/setting/bank-account/bank-account.service.ts
```diff
@@ -0,0 +1,12 @@
+import { BaseService } from '@/app/api/core/services/base.service'
+import { QBAccountQueryResponseSchema } from '@/type/dto/intuitAPI.dto'
+import IntuitAPI, { QB_ACCOUNT_COLUMNS } from '@/utils/intuitAPI'
+
+export class BankAccountService extends BaseService {
+  async listActiveBankAccounts(intuitApi: IntuitAPI) {
+    const rawResult = await intuitApi.customQuery(
+      `SELECT ${QB_ACCOUNT_COLUMNS.join(', ')} FROM Account WHERE AccountType = 'Bank' AND Active = true maxresults 100`,
+    )
```
priosshrsth, line 11: @SandipBajracharya Is it supposed to list all active bank accounts? In which case, isn't it false to say these are all acgtive bank accounts when we list max of 100?

src/app/api/quickbooks/setting/setting.controller.ts
```diff
@@ -22,12 +26,23 @@ export async function getSettings(req: NextRequest) {
...
+        'bankDepositFeeFlag',
+        'useCompanyNameFlag',
+      )
     if (parsedType.data === SettingType.PRODUCT)
       returningFields.push('createNewProductFlag')
   }
   const setting = await settingService.getOneByPortalId(returningFields)
-  return NextResponse.json({ setting })
+
+  let bankAccountRef: string | null = null
```
priosshrsth: Let's avoid `let`

src/type/common.ts
```diff
@@ -288,6 +290,21 @@ export const SettingRequestSchema = z
           message: 'absorbedFeeFlag is required when type is invoice',
         })
       }
+      if (typeof val.bankDepositFeeFlag !== 'boolean') {
```
priosshrsth, line 293: @SandipBajracharya So if type is invoice, then bankDepositFreeFlag is required? DO you think this coul dbe simpleified by union type?

z.union([SettingSChema, InvoiceSettingSchema]) ... or something like this?

No need to spend much time on it if it bring type issue and new challenges.

src/hook/useSettings.ts
priosshrsth: @SandipBajracharya setAppParams dependency issue. Can you suppress this?

priosshrsth (changes requested):

src/type/common.ts
SandipBajracharya: There were few type issues so deferring this for now. Nice suggestion though. Thanks

src/app/api/quickbooks/setting/bank-account/bank-account.service.ts
SandipBajracharya: increase maxResult to 1000. I think that is reasonable as its rare case for a company to have more than 1k active bank accounts.

priosshrsth (approved):
lgtm 🏆

PR #270, opened 2026-07-27
OUT-4010: freeze batched-deposit intent per invoice

SandipBajracharya:
## OUT-4010 — Freeze batched-deposit intent per invoice

https://linear.app/assemblycom/issue/OUT-4010

### Problem
`bankDepositFeeFlag` was read **live** in both `handlePaymentSucceeded` and `handlePayoutReconciliationCompleted`. Toggling it between a payment and its payout desynced the two paths — the fee got booked twice (OFF→ON) or missed while the payment stranded in Undeposited Funds (ON→OFF).

### Approach
Freeze the decision per invoice. A new `qb_invoice_sync.is_batched_deposit` column is set from the live flag at invoice-create time; both handlers read the frozen value.

- `resolveDepositToAccountRef` takes the frozen flag; the only live read stays at the freeze point (create / paid-on-create).
- `payment.succeeded`: dedupe + resolve the frozen intent before the claim; a batched invoice defers to the payout with zero rows.
- `payout`: resolves per-invoice intent via `getSuccessfulPaidPaymentIds` (now joined to `qb_invoice_sync`); all-batched books one deposit, all-non-batched skips before claiming, mixed is rejected.

### Tests
- New freeze/routing coverage: `freezeBatchedIntent`, `statusPaidDepositRouting`, `frozenIntentRouting`, `frozenIntentDefer`, plus payout `allNonBatched` / `mixed`.
- Reconciled the OUT-4006 payout suite with the frozen model: ported edge-case tests to `seedPaidInvoiceForPayout`, updated `resolvePayments` for the new return shape, removed the obsolete live-flag tests, and folded the duplicate happy-path/parallel infra into one directory.
- Full suite green: **350 passing**.

### Notes for reviewers
- Migration is a plain additive column (default `false`); no backfill needed since the batched-deposit feature isn't in production yet, so no pre-existing invoice was created under batched behavior.


SandipBajracharya:
@greptileai review PR again

src/app/api/quickbooks/webhook/webhook.service.ts
```diff
@@ -488,102 +513,121 @@ export class WebhookService extends BaseService {
...
+    const syncLogService = new SyncLogService(this.user)
+    // Cheap duplicate short-circuit: a redelivered event already has a
+    // SUCCEEDED claim row, so skip the sleep + Copilot fetch. The atomic
+    // claim below still guards the first-delivery race.
+    const existingPaymentLog =
+      await syncLogService.getOneByCopilotIdAndEventType({
+        copilotId: paymentId,
         eventType: EventType.SUCCEEDED,
         entityType: EntityType.PAYMENT,
       })
```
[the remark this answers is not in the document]
SandipBajracharya: Intentional — the peek is status-blind to mirror claimWebhookEvent's onConflictDoNothing, which already blocks any redelivery regardless of status. The webhook returns 200 even on failure, so Copilot never redelivers a FAILED event as a retry; recovering a failed attempt is the resync cron's job, not the webhook's. Reworded the comment/log in e19d860 to make that explicit.

src/app/api/quickbooks/webhook/webhook.service.ts
```diff
@@ -653,33 +697,45 @@ export class WebhookService extends BaseService {
...
           httpStatus.NOT_FOUND,
-          `Payout ${payout.id}: no SUCCESS INVOICE/PAID sync log for invoices [${unresolved.join(', ')}]`,
+          `Payout ${payoutId}: no SUCCESS INVOICE/PAID sync log for invoices [${unresolved.join(', ')}]`,
+        )
+      }
+
+      // All-non-batched skipped pre-claim; a non-batched invoice here = mixed.
+      // get() is non-null — every id passed the unresolved check above.
+      const allBatched = copilotInvoiceIds.every(
+        (id) => paymentIdByInvoice.get(id)!.isBatchedDeposit,
```
priosshrsth: @SandipBajracharya I think better to use `?` rather than `!` imo.

priosshrsth (changes requested):

src/app/api/quickbooks/webhook/webhook.service.ts
SandipBajracharya: Converted to optional chaining

priosshrsth (approved):
lgtm

PR #271, opened 2026-07-28
OUT-4011: Name affected invoices in mixed-payout reconciliation notification

SandipBajracharya:
## Summary

When a Stripe payout mixes invoices set to batch into a bank deposit with invoices that are not, no deposit can be booked and the payout needs manual reconciliation. This surfaces that case to internal users as a `QB_PAYOUT_MIXED_INTENT` notification, and — the focus of this branch — **names the specific invoice numbers** in the notification body so it's clear which invoices went unrecorded.

## What changed

- **Notification declarations** — `PAYOUT_MIXED_INTENT_CODE` sentinel + `AppActionableErrorCodes` routing, an `invoiceNumbers` context field, and the `QB_PAYOUT_MIXED_INTENT` in-product/email copy that lists the affected invoices.
- **Service logic** — the webhook detects mixed intent (`MixedPayoutIntentError`), writes a single terminal FAILED sync-log row tagged with the sentinel, and stashes the affected invoice numbers in `remark` (a payout spans multiple invoices, so they don't fit the single `invoice_number` column — no schema change). `getSuccessfulPaidPaymentIds` returns each invoice number; `SyncErrorNotifier` surfaces the list while keeping the payout id as the entity reference.

## Design notes

- No DB migration: `remark` is an existing free-text column; only the value written for the mixed-payout FAILED case changed.
- The payout id stays as the notification "ref" for support traceability; the invoice list is an added clause, not a replacement.

## Testing

- New/updated coverage: mixed-payout persists invoice numbers to `remark`, `resolvePayments` returns them, the notifier renders the invoice list through the real copy (both channels), and a regression test for `invoice.voided` on an OPEN row surfacing a QBO **5010** as `error_code` → `QB_STALE_OBJECT`.
- Full suite green: **64 unit + 101 integration** tests pass.

### DIFF
--- a/src/app/api/notification/notification.helper.ts
+++ b/src/app/api/notification/notification.helper.ts
@@ -249,14 +249,14 @@ export const NotificationCopy: Record<
       const forInvoices = ctx?.invoiceNumbers
         ? ` for invoices ${ctx.invoiceNumbers}`
         : ''
-      return `A Stripe payout${ref} could not be recorded in QuickBooks because it mixes invoices set to batch into a bank deposit with invoices that are not — this happens when the bank-deposit setting changed between a payment and its payout. No deposit was created${forInvoices}, so nothing was double-booked. Record this payout's deposit manually in QuickBooks. This will not retry automatically.`
+      return `A Stripe payout${ref} could not be recorded in QuickBooks because it mixes invoices set to batch into a bank deposit with invoices that are not — this happens when the bank-deposit setting changed between a payment and its payout. No deposit was created${forInvoices}, so nothing was double-booked. Record this payout's deposit manually in QuickBooks. This will not be retried automatically.`
     },
     emailSubject: 'QuickBooks sync failed: payout needs manual reconciliation',
     emailBody: (ref, ctx) => {
       const forInvoices = ctx?.invoiceNumbers
         ? ` for invoices ${ctx.invoiceNumbers}`
         : ''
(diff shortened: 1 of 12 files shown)


src/app/api/notification/notification.helper.ts
```diff
@@ -236,6 +239,27 @@ export const NotificationCopy: Record<
...
+  // both batched and non-batched invoices. We can't book one balanced deposit
+  // from a mix, so no deposit is created and the payout needs manual handling.
+  // Terminal — there is no scheduled retry for payouts.
+  [NotificationActions.QB_PAYOUT_MIXED_INTENT]: {
+    title: 'QuickBooks sync failed: payout needs manual reconciliation',
+    body: (ref, ctx) => {
+      const forInvoices = ctx?.invoiceNumbers
+        ? ` for invoices ${ctx.invoiceNumbers}`
+        : ''
+      return `A Stripe payout${ref} could not be recorded in QuickBooks because it mixes invoices set to batch into a bank deposit with invoices that are not — this happens when the bank-deposit setting changed between a payment and its payout. No deposit was created${forInvoices}, so nothing was double-booked. Record this payout's deposit manually in QuickBooks. This will not retry automatically.`
```
priosshrsth: `not be retried automatically` might make more sense?

src/app/api/quickbooks/syncLog/syncErrorNotifier.ts
priosshrsth, line 72: Maybe be just make the underfined attribute as nullable in type? Most of the time it doesn't need any change since both null and undefined act as falsy/empty value. Then we can avoid ?? undefined.

priosshrsth (changes requested):

SandipBajracharya:
@greptileai review PR again

priosshrsth (approved):
lgtm

PR #272, opened 2026-07-29
OUT-4012: warn before changing the bank-deposit flag

SandipBajracharya:
## Problem

Changing `bankDepositFeeFlag` has consequences users can't see: the deposit behavior is frozen per invoice at creation, so a toggle only affects new invoices, and a Stripe payout that mixes pre/post-change invoices won't get its automatic bank deposit and needs manual reconciliation.

## Shipped

A confirmation modal on **save** (not on toggle), shown only when the flag differs from its saved value — covers both enable and disable, and never fires when unchanged. UX safeguard only; correctness lives in the frozen-intent + mixed-guard work (OUT-4010/4009).

Copy:
> **Change bank deposit setting?**
> This applies only to invoices created from now on; existing invoices are unaffected. If a Stripe payout mixes invoices from before and after the change, you'll need to reconcile that payout manually in QuickBooks.

## Code sites

- `src/components/ui/ConfirmModal.tsx` *(new)* — reusable portal modal (Escape/backdrop dismiss, `role="dialog"` + `aria-labelledby`/`aria-describedby` via `useId`).
- `src/hook/useSettings.ts` — `useInvoiceDetailSettings` gains the save-intercept (`requestInvoiceSettingsSave` / `confirmBankDepositChange` / `cancelBankDepositChange`); raw `submitInvoiceSettings` no longer exported.
- `src/components/dashboard/settings/SettingAccordion.tsx` — invoice save button routes through the guard and renders the modal. `InvoiceDetail.tsx` unchanged (checkbox stays a free local toggle).

## Notes

- Stacked on **OUT-4011** (base branch), not master.
- No component-test infra exists in the repo (both Vitest projects run `environment: 'node'`); verified via `tsc`, lint, prettier, `yarn build`, and `yarn test` (357/357).
- Known follow-up: on a brand-new portal's first-time setup, enabling the flag trips the modal (saved value genuinely changed) — cosmetic, suppression deferred pending a product call.

## Testing Criteria

image

### DIFF
--- a/src/components/ui/ConfirmModal.tsx
+++ b/src/components/ui/ConfirmModal.tsx
@@ -25,24 +25,28 @@ export default function ConfirmModal({
   const titleId = useId()
   const descId = useId()
   const dialogRef = useRef<HTMLDivElement>(null)
-  // Keep the latest onCancel without re-running the focus effect each render.
-  const onCancelRef = useRef(onCancel)
-  onCancelRef.current = onCancel
 
-  // On open: focus into the dialog, trap Tab, and restore focus on close.
+  // On open, focus into the dialog; on close, restore focus to the opener.
   useEffect(() => {
     if (!open) return
     const previouslyFocused = document.activeElement as HTMLElement | null
-    const focusables = Array.from(
-      dialogRef.current?.querySelectorAll<HTMLElement>('button') ?? [],
-    )
-    focusables[0]?.focus()
+    const buttons = dialogRef.current?.querySelectorAll<HTMLElement>('button')
+    buttons?.[0]?.focus()
+    return () => previouslyFocused?.focus()
+  }, [open])
 
+  // Escape cancels; Tab is trapped between the dialog's buttons.
+  useEffect(() => {
+    if (!open) return
     const onKeyDown = (e: KeyboardEvent) => {
-      if (e.key === 'Escape') return onCancelRef.current()
-      if (e.key !== 'Tab' || focusables.length === 0) return
-      const first = focusables[0]
-      const last = focusables[focusables.length - 1]
+      if (e.key === 'Escape') return onCancel()
+      if (e.key !== 'Tab') return
+      const buttons = Array.from(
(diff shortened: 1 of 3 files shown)


SandipBajracharya:
@greptileai review PR again

src/components/ui/ConfirmModal.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  if (!open) return null
+
+  return createPortal(
+    <div
+      className="fixed inset-0 z-50 flex items-center justify-center bg-black/40"
+      onClick={onCancel}
+    >
+      <div
+        role="dialog"
```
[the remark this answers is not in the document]
SandipBajracharya: Fixed in e12f849. Now when the modal opens, focus jumps into it and stays trapped inside (Tab can't reach the buttons behind it), and when it closes focus goes back to where it was.

src/components/ui/ConfirmModal.tsx
```diff
@@ -0,0 +1,91 @@
...
+  onConfirm,
+  onCancel,
+}: ConfirmModalProps) {
+  const titleId = useId()
+  const descId = useId()
+  const dialogRef = useRef<HTMLDivElement>(null)
+  // Keep the latest onCancel without re-running the focus effect each render.
+  const onCancelRef = useRef(onCancel)
+  onCancelRef.current = onCancel
+
```
priosshrsth: @SandipBajracharya I am unable to understand how this helps to prevent unnecessary focus? maybe you sould use two useEffects?

priosshrsth (changes requested):

src/components/ui/ConfirmModal.tsx
SandipBajracharya: Good call, split it into two. One runs on open for focus + restore, the other handles Escape/Tab. Also dropped the ref.

priosshrsth (approved):
lgtm

PR #274, opened 2026-07-31
feat(OUT-4005): payout reconciliation service + deposit lookup [2/4]

SandipBajracharya:
**Stack 2/4** — base: `OUT-4005-split-1`

`PayoutService.reconcile` validates the payout, resolves its payments, and builds one batched QBO deposit. On resync it reuses an already-made deposit (stored id, then a txn-date query matching `PrivateNote`) so a retry can't create a duplicate. Adds `getDepositsByTxnDate` + schemas, error types, and unit/service tests.

Part of the OUT-4005 payout-reconciliation stack:
1. schema + migration
2. **reconciliation service + deposit lookup** ← this PR
3. webhook + resync wiring
4. typecheck tooling + chore

### DIFF
--- a/src/app/api/quickbooks/payout/payout.service.ts
+++ b/src/app/api/quickbooks/payout/payout.service.ts
@@ -1,0 +1,225 @@
+import httpStatus from 'http-status'
+import { and, eq, isNull } from 'drizzle-orm'
+
+import { BaseService } from '@/app/api/core/services/base.service'
+import APIError from '@/app/api/core/exceptions/api'
+import { SyncLogService } from '@/app/api/quickbooks/syncLog/syncLog.service'
+import { PaymentService } from '@/app/api/quickbooks/payment/payment.service'
+import { TokenService } from '@/app/api/quickbooks/token/token.service'
+import {
+  MixedPayoutIntentError,
+  TerminalPayoutError,
+} from '@/app/api/quickbooks/payout/payout.errors'
+import {
+  QBPayoutSync,
+  QBPayoutSyncSelectSchemaType,
+} from '@/db/schema/qbPayoutSync'
+import { PayoutLineItem } from '@/type/dto/webhook.dto'
+import IntuitAPI, { IntuitAPITokensType } from '@/utils/intuitAPI'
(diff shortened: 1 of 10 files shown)


src/app/api/quickbooks/payout/payout.service.ts
```diff
@@ -0,0 +1,225 @@
...
+  QBPayoutSync,
+  QBPayoutSyncSelectSchemaType,
+} from '@/db/schema/qbPayoutSync'
+import { PayoutLineItem } from '@/type/dto/webhook.dto'
+import IntuitAPI, { IntuitAPITokensType } from '@/utils/intuitAPI'
+import { AccountTypeObj } from '@/constant/qbConnection'
+import { validateAccessToken } from '@/utils/auth'
+
+export class PayoutService extends BaseService {
+  private syncLogService = new SyncLogService(this.user)
```
priosshrsth: Does this work? what would this be. Can you maybe move this to constructor instead?

src/utils/intuitAPI.ts
```diff
@@ -1134,5 +1171,6 @@ export default class IntuitAPI {
   deletePayment = this.wrapWithRetry(this._deletePayment)
   deletePurchase = this.wrapWithRetry(this._deletePurchase)
   createDeposit = this.wrapWithRetry(this._createDeposit)
+  getDepositsByTxnDate = this._getDepositsByTxnDate.bind(this)
```
priosshrsth, line 1174: What is the benifit of doing it this way?

priosshrsth (changes requested):

src/app/api/quickbooks/payout/payout.service.ts
SandipBajracharya: Yes it works, but the constructor should be good for this case.

src/utils/intuitAPI.ts
SandipBajracharya: `getDepositsByTxnDate` is not wrapped with retry. In fact we do not wrap get functions with retry.

priosshrsth (approved):
lgtm

PR #280, opened 2026-08-07
OUT-4030: warn IUs which mixed-payout fees are already recorded

SandipBajracharya:
## Problem

A Stripe payout can straddle a `bankDepositFeeFlag` toggle, mixing invoices set to batch into a bank deposit with invoices that are not. These payouts are unsupported and fail terminally (FAILED `PAYOUT/SETTLED`, no auto-retry) with an IU notification.

The old notification told the IU to record the deposit manually and stated nothing was double-booked. But the non-batched invoices already have their absorbed Stripe fee posted as a QBO expense. Following that guidance — booking a normal batched deposit (gross lines minus a fee line) — **double-books** those fees.

## Change

Keep mixed payouts terminal + manual (no auto delete/rebuild — the fee-funding account `assetAccountRef` is user-selectable to a real, reconciled bank, so deleting posted fee expenses isn't safe). Instead, make the notification actionable:

- **Ground-truth lookup** (`getInvoiceNumbersWithRecordedFee`): invoices that have a SUCCESS `PAYMENT/SUCCEEDED` sync log — the fee Purchase exists in QBO only when that row is written, so it's ground truth, not inferred from the batched-intent flag. Standalone `@/db/service` fn to avoid a `syncLog.service ↔ syncErrorNotifier` import cycle.
- **Notifier** derives the list from the payout's affected invoices and populates `NotificationContext.invoiceNumbersWithFee`. A lookup failure drops only that detail — the terminal payout notification still fires.
- **Copy** (in-product + email) names the already-recorded invoices and warns not to record those fees again.
- Shared `MIXED_INTENT_INVOICE_DELIMITER` so the remark writer and the notifier split can't drift.

## Testing

- New integration test exercises the query against real Postgres: SUCCESS included; FAILED / wrong-entity / wrong-event / soft-deleted / other-portal excluded; empty input → empty set.
- Unit tests: copy renders the warning when fees are recorded and omits it otherwise; notifier populates the field, preserves remark order, and still dispatches when the lookup throws.
- Full suite green (407 tests); tsc/eslint/prettier clean for all files in this PR.

Sub-issue of OUT-3604. Linear: https://linear.app/assemblycom/issue/OUT-4030

### DIFF
--- a/src/app/api/core/types/notification.ts
+++ b/src/app/api/core/types/notification.ts
@@ -31,6 +31,9 @@ export interface NotificationContext {
   // Comma-joined invoice numbers for a multi-invoice failure (mixed payout),
   // where the single invoiceNumber above can't hold them all.
   invoiceNumbers?: string | null
+  // Subset of invoiceNumbers whose absorbed fee is already recorded in QBO, so a
+  // mixed-payout body can tell IUs which fees not to record a second time.
+  invoiceNumbersWithFee?: string | null
   customerName?: string | null
   productName?: string | null
   qbItemName?: string | null
--- a/src/app/api/quickbooks/syncLog/syncErrorNotifier.ts
+++ b/src/app/api/quickbooks/syncLog/syncErrorNotifier.ts
@@ -7,10 +7,13 @@ import {
 import { NotificationService } from '@/app/api/notification/notification.service'
 import {
   AppActionableErrorCodes,
+  MIXED_INTENT_INVOICE_DELIMITER,
   UserActionableErrorCodes,
 } from '@/constant/intuitErrorCode'
 import { QBSyncLogSelectSchemaType } from '@/db/schema/qbSyncLogs'
 import { getPortalConnection } from '@/db/service/token.service'
+import { getInvoiceNumbersWithRecordedFee } from '@/db/service/syncLog.service'
+import CustomLogger from '@/utils/logger'
 
 /**
  * Looks up the user-actionable notification action for a given QBO error code.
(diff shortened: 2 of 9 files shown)


src/app/api/quickbooks/syncLog/syncErrorNotifier.ts
```diff
@@ -69,6 +72,35 @@ export class SyncErrorNotifier extends BaseService {
...
+    // Mixed-payout rows stash the affected invoices in `remark`; surface them,
+    // then flag which already have a recorded fee to warn against double-booking.
+    const affectedInvoiceNumbers =
+      action === NotificationActions.QB_PAYOUT_MIXED_INTENT ? log.remark : null
+    let invoiceNumbersWithFee: string | undefined
+    if (affectedInvoiceNumbers) {
+      try {
+        const affected = affectedInvoiceNumbers
+          .split(MIXED_INTENT_INVOICE_DELIMITER)
+          .filter(Boolean)
```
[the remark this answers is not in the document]
SandipBajracharya: Invoice number cannot have comma and space. This case won't come to be realistic.

SandipBajracharya:
@greptileai review PR again

src/app/api/core/types/notification.ts
```diff
@@ -31,6 +31,9 @@ export interface NotificationContext {
   // Comma-joined invoice numbers for a multi-invoice failure (mixed payout),
   // where the single invoiceNumber above can't hold them all.
   invoiceNumbers?: string | null
+  // Subset of invoiceNumbers whose absorbed fee is already recorded in QBO, so a
+  // mixed-payout body can tell IUs which fees not to record a second time.
+  invoiceNumbersWithFee?: string | null
```
priosshrsth, line 36: why string instead of string[].

It suggest that this is going to be plural.

Ok. Looks like this is intended to be comma separated string. But why this pattern instead of using array?

src/app/api/quickbooks/syncLog/syncErrorNotifier.ts
```diff
@@ -69,6 +72,35 @@ export class SyncErrorNotifier extends BaseService {
...
+          invoiceNumbersWithFee = recorded.join(MIXED_INTENT_INVOICE_DELIMITER)
+      } catch (error) {
+        // A lookup blip must still let the terminal payout notification through.
+        CustomLogger.error({
+          message:
+            'SyncErrorNotifier | recorded-fee lookup failed; notifying without it',
+          obj: error,
+        })
+      }
+    }
```
priosshrsth: `const {affectedInvoceNumbers, invoiceNumbersWithFee} = action !== NotificationActions.QB_PAYOUT_MIXED_INTENT ? {} : someFunc(..params)`

I'd suggest doing somethin like this.

priosshrsth (changes requested):

src/app/api/core/types/notification.ts
SandipBajracharya: Since we need to show those invoice numbers in notification body, I preferred to use string instead. What I did is basically `join()` the array of string from the call sites.

priosshrsth (approved):
lgtm. 🏆

